PR #3, opened 2018-09-20
Raymond Shiner - Assignment 1 Submission

raymondshiner:
Made a new file to test commits and pushes as shown in class.


Assignment1/test/Assignment1.Tests/UnitTest1.cs
```diff
@@ -0,0 +1,44 @@
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using System;
+
+namespace Assignment1.Tests
+{
+    [TestClass]
```
breaman: Nice work on having multiple unit tests to make sure that things behave the way you really expected them to.

PR #4, opened 2018-09-20
My first committed repository.

NikolasGraham:
This is a test for pull request (CAN BE IGNORED)


Assignment1/HelloWorld/test/HelloWorld.Tests/UnitTest1.cs
```diff
@@ -0,0 +1,19 @@
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace HelloWorld.Tests
+{
+    [TestClass]
+    public class UnitTestName
```
breaman: While you did rename the class, it would have been better to call this UserInputNameTests since that is the class it is testing.

Assignment1/HelloWorld/test/HelloWorldTests/HelloWorldTests.csproj
```diff
@@ -0,0 +1,20 @@
+<Project Sdk="Microsoft.NET.Sdk">
```
breaman: Looks like this project was an extra that didn't get deleted. Not docking for it, but might want to make sure you only check in code that you intend to have run/included.

PR #5, opened 2018-09-20
Added the submission of my Assignment1

KennethWhite:
Also made changes to the .gitignore file ignoring some VS project files


Assignment1/HelloWorld/HelloWorld.Tests/UnitTest1.cs
```diff
@@ -0,0 +1,19 @@
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using System;
+
+namespace HelloWorld.Tests
+{
+    [TestClass]
+    public class UnitTest1
```
breaman: Probably want to make the class name a little more specific. In this case you are testing the Program class, so something like ProgramTests would be better.

PR #7, opened 2018-09-20
My first push

bcaesar42:


Assignment1/HelloWorld/src/HelloWorld/HelloWorld.csproj
```diff
@@ -0,0 +1,8 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>netcoreapp2.2</TargetFramework>
```
breaman: Looks like you installed the latest preview bits on your machine since you are targeting netcoreapp2.2. While I have that installed on my machine as well, I don't believe the lab machines have that version installed and I don't believe the other instructors have that installed, so this application version might not work on other machines. Just an fyi.

PR #11, opened 2018-09-20
Added my text file

Nematodes:
Accidentally added .vs to my fork, so I had to figure out how to delete it. Now the only new thing is the text file.


Assignment1/HelloSandwich/HelloSandwich/HelloSandwich.cs
```diff
@@ -0,0 +1,30 @@
+﻿using System;
+
+/*
+ * HelloSandwich.cs
+ * Created by Brian Bos
+ * 
+ * I am using \n instead of System.Environment.NewLine here.
```
breaman: Out of curiosity, why use \n instead?

Assignment1/HelloSandwich/.gitignore
```diff
@@ -0,0 +1 @@
+.vs
```
breaman: This .gitignore file let a lot more files go into github than what should have been allowed. This assignment should have had about 6-10 files checked in, not necessarily 57. For future projects, you will want a more robust .gitignore file. Having all these files in the PR made it more confusing than it needed to be in order to review what was actually needed.

Assignment1/HelloSandwich/HelloSandwichUnitTest/UnitTest1.cs
```diff
@@ -0,0 +1,49 @@
...
+ * (Unit test output end)
+ */
+
+namespace BrianBosAssignment1Namespace
+{
+    [TestClass]
+    public class HelloSandwichUnitTester
+    {
+        [TestMethod]
+        public void TestHelloSandwich()
```
breaman: This was great to have multiple tests, but why comment them out? Why not run all of them or break them out into separate test methods? Not marking down for this, but no point in putting in code and then commenting it (especially when it is verifying that your code is behaving correctly).

Assignment1/HelloSandwich/HelloSandwich/HelloSandwich.cs
Nematodes: Using \n over System.Environment.Newline was due to its ease of use. System.Environment.Newline is obviously more portable, but involves marginally more effort to utilize than \n, and the project was very simple to begin with.

My next assignment (assignment 2) no longer uses \n for additional line breaks, but interpolates System.Environment.Newline into the strings instead.

PR #13, opened 2018-09-25
My pull request

JeffHowes:
First pull.


Assignment1/test/HelloWorld.test/UnitTest1.cs
```diff
@@ -0,0 +1,27 @@
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace HelloWorld.test
+{
+    [TestClass]
+    public class HelloWorld_UnitTest1  
+    {
+        [TestMethod]
+        public void Validat_HelloWorld_TestMethod()
```
breaman: Careful of typo's here. My guess is you wanted "Validate". Not marking you down, but I did have a teacher that would have :)

PR #15, opened 2018-09-27
Assignment 1 Pull Request

blazinaj:
9.27.18


Assignment1/test/Assignment1.Tests/Assignment1.Test.cs
```diff
@@ -0,0 +1,18 @@
...
+namespace Assignment1.Tests
+{
+    [TestClass]
+    public class InputOutputTest
+    {
+        [TestMethod]
+        public void TestInputOutput()
+        {
+            string expectedOutput = $">>Hello, what is your name?\n" +
+                "<<{expectedInput}\n" +
```
breaman: Although this works, my guess is this was not what your expected behavior was. You are passing in the string "{expectedInput}" into the console instead of an actual value. I have a feeling you wanted to declare a variable named expectedInput and then used that. In order to do that, this line needed to be $"<<{expectedInput}". Also, to be safe in all environments (and it wasn't covered in class, so just pointing this out for future reference) you should have used Environment.NewLine instead of \n.

Assignment1/test/Assignment1.Tests/Assignment1.Test.cs
```diff
@@ -0,0 +1,18 @@
...
+                ">>Hello {expectedInput}!\n";
+
+            IntelliTect.TestTools.Console.ConsoleAssert.Expect(expectedOutput, global::Assignment1.Assignment1.Main);
```
breaman: While "global::" works, you don't need it in this context.

PR #17, opened 2018-09-27
Assignment1

toha87:
Assignment1


Assignment1/Assignment 1/Program.cs
```diff
@@ -0,0 +1,14 @@
+﻿using System;
+
+namespace Assignment_1
+{
+    public class ProgramInputOutput
+    {
+        public static void Main(string[] args)
+        {
+            Console.WriteLine("Hello, whats your name?");
```
breaman: Not grading down for spelling errors (although I had a teacher that did that), you might still want to make sure you don't have typo's/spelling errors in your strings (should have been "what's").

Assignment1/Assignment1UnitTest/UnitTest1.cs
```diff
@@ -0,0 +1,21 @@
+﻿using System;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace Assignment1UnitTest
+{
+    [TestClass]
+    public class UnitTestClassForAss1
+    {
+        [TestMethod]
+        public void TestMethodInputOutput()
```
breaman: Nice work on this test. While I didn't cover it in class, in the future for something like this, you might have wanted to have a second unit test to make sure that the behavior was actually what you were expecting. Technically your program could have been hardcoded to return "Hello Anatoli" and while this test would have passed, a test that had a different string as input would have then failed.

PR #26, opened 2018-10-04
Assignment2

KennethWhite:
Assignment 2 solution and unit tests.


Assignment2/Assignment2/Assignment2.UnitTests/Assignment2.UnitTests.cs
```diff
@@ -0,0 +1,411 @@
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using System;
+using System.Text;
+
+namespace Assignment2.UnitTests
+{
+
+    [TestClass]
+    public class StringUnitTests
```
Keboo: In the future break out each `class` into its own file.

Assignment2/Assignment2/Assignment2.UnitTests/Assignment2.UnitTests.cs
```diff
@@ -0,0 +1,411 @@
...
+        }
+
+        [TestMethod]
+        public void TestStringCompareTo()
+        {
+            string fruit = "Apples";
+            string veggie = "Squash";
+            Assert.IsTrue(fruit.CompareTo(veggie) < 0) ;
+            Assert.IsTrue(veggie.CompareTo(fruit) > 0);
+            Assert.IsTrue(fruit.CompareTo(fruit) == 0);
```
Keboo: Great job testing all three cases, though typically each case would be its own unit test (or use DataRows like you did below).

Assignment2/Assignment2/Assignment2.UnitTests/Assignment2.UnitTests.cs
```diff
@@ -0,0 +1,411 @@
...
+        }
+        [DataRow("Bonanza", "Bon", true)]
+        [DataRow("Crabapples", "dog", false)]
+        [DataRow("Micrsoft", "soft", true)]
+        [DataRow("wow", "wow", true)]
+        [TestMethod]
```
Keboo: Great use of data driven tests!

Assignment2/Assignment2/Assignment2.UnitTests/Assignment2.UnitTests.cs
```diff
@@ -0,0 +1,411 @@
...
+        }
+
+        [DataRow("", true)]
+        [DataRow(" ", false)]
+        [DataRow(null, true)]
+        [DataRow("Something", false)]
+        [TestMethod]
+        public void TestIsNullOrEmpty(string word, bool result)
+        {
+            Assert.AreEqual(String.IsNullOrEmpty(word), result);
```
Keboo: In general favor using C# keywords rather than type names. So `string.IsNullOrEmpty` rather than `String.IsNullOrEmpty`

Assignment2/Assignment2/Assignment2.UnitTests/Assignment2.UnitTests.cs
```diff
@@ -0,0 +1,411 @@
...
+        }
+
+    }//Output/Exit Unit tests
+
+    [TestClass]
+    public class EdgeCaseTests
+    {
+
+        [TestMethod]
+        public void EmptyLine()
```
Keboo: In general you want your unit test names to be a bit more explicit. I would have liked to see something indicating what the empty line was expected to produce as part of the test name.

Assignment2/Assignment2/Assignment2.UnitTests/Assignment2.UnitTests.cs
```diff
@@ -0,0 +1,411 @@
...
+Enter binomial math operation to solve. e.g. (-2.5%5.1): <<{line}
+>>Enter binomial math operation to solve. e.g. (-2.5%5.1): <<{exitLine}
+>>Program complete.";
+            IntelliTect.TestTools.Console.ConsoleAssert.Expect(expectedOutput, StringMath.StringMath.Main);
+        }
+
+        [TestMethod]
+        public void AddIntMinMaxValue()
+        {
+            StringBuilder sb = new StringBuilder(100);
```
Keboo: If you are only going to have a single call to `.Append` on your `StringBuilder` it is better to just use a string directly.

Assignment2/Assignment2/Assignment2/StringMath.cs
```diff
@@ -0,0 +1,133 @@
...
+            if (args.Length == 1)
+                argsFlag = true;
+          
+
+            Console.WriteLine("Enter 'exit' to end program.");
+            bool runFlag = true;
+
+            while (runFlag) {
+                Console.Write("Enter binomial math operation to solve. e.g. (-2.5%5.1): ");
+                line = (argsFlag)? args[0] : Console.ReadLine();
```
Keboo: Good use of ternary operator!

Assignment2/Assignment2/Assignment2/StringMath.cs
```diff
@@ -0,0 +1,133 @@
...
+                        return Math.Pow(lhs, rhs);
+                default:
+                    throw new ArgumentException("Unsupported operator");
+            }
+       
+        }
+
+        private static void PrintFormattedResult(MatchCollection operands,char op, double result)
+        {
+            Console.WriteLine($"{operands[0]}{op}{operands[1]} = {String.Format("{0:0.####}", result)}");
```
Keboo: When inside of an interpolated string you don't need to use `string.Format` instead you can specify the format string directly.
```C#
Console.WriteLine($"{operands[0]}{op}{operands[1]} = {result:0.####}");
```

Assignment2/Assignment2/Assignment2/StringMath.cs
```diff
@@ -0,0 +1,133 @@
...
+            switch (op)
+            {
+                case '+':
+                    return lhs + rhs;
+                case '-':
+                    return lhs - rhs;
+                case '*':
+                    return lhs * rhs;
+                case '/':
+                    if (Math.Abs(rhs) < Double.Epsilon)
```
Keboo: Good use of double comparison!

Assignment2/Assignment2/Assignment2/StringMath.cs
```diff
@@ -0,0 +1,133 @@
...
+            double lhs, rhs;
+            if (lDec || rDec)
+            {
+                Double.TryParse(lhsString, out lhs);
+                Double.TryParse(rhsString, out rhs);
+            }
+            else
+            {
+                long.TryParse(lhsString, out long lTemp);
+                long.TryParse(rhsString, out long rTemp);
```
Keboo: You probably wanted to use `double.TryParse` for everything. Switching to longs, reduces the number of values you can handle. Consider that "9223372036854775808" (long.MaxValue + 1), can be parsed as a double.

Assignment2/Assignment2/Assignment2/StringMath.cs
```diff
@@ -0,0 +1,133 @@
...
+                PrintFormattedResult(operands, op, result);
+            }
+            return runFlag;
+        }
+
+        private static MatchCollection ParseLineParameters(string line)
+        {
+            line = System.Text.RegularExpressions.Regex.Replace(line, " ", "");
+            //0-1 minus signs followed by 1+ digits, optional decimal point and precision digits
+            Regex regex = new Regex(@"(?<!\d)-?(\d+)?(\.\d+)?");
```
Keboo: Nice use of regex to parse out the numbers!

Keboo:
Really nice work!

PR #27, opened 2018-10-04
Raymond Shiner - Assignment2 Submission

raymondshiner:


Assignment2/src/Assignment2/Program.cs
```diff
@@ -0,0 +1,134 @@
...
+
+                        num1 = num1 * (-1);         //make num1 negative
+                    }
+
+                    else                            // second num negative - first number postiive - this splits as such: "2--3" --> { "2", "", "3" }
+                    {
+                        int.TryParse(array[0], out num1);
+                        int.TryParse(array[2], out num2);
+
+                        num2 = num2 * (-1);         //make num2 negative
```
Keboo: This can be simplified by doing: `num2 = -num2;`

Assignment2/test/Assignment2.Tests/StringTests.cs
```diff
@@ -0,0 +1,43 @@
...
+    public class StringTests
+    {
+        [TestMethod]
+        public void TestStringEquals()
+        { 
+            Assert.AreEqual(true, "Raymond".Equals("Raymond"));
+            Assert.AreEqual(true, "Frank".Equals("Frank"));
+            Assert.AreEqual(false, "One".Equals("Two"));
+            Assert.AreEqual(true, "Three".Equals("Three"));
+        }
```
Keboo: There is also `Assert.IsTrue` and `Assert.IsFalse` that can make some of these simple boolean comparisons easier.

Assignment2/test/Assignment2.Tests/StringTests.cs
```diff
@@ -0,0 +1,43 @@
...
+            Assert.AreEqual(3, "123".Length);
+        }
+
+        [TestMethod]
+        public void TestStringTrim()
+        {
+            Assert.AreEqual("Trim", " Trim  ".Trim());
+            Assert.AreEqual("Heyo", " Heyo".Trim());
+            Assert.AreEqual("This", "This ".Trim());
+        }
```
Keboo: Trim also strips whitespace from the beginning of the string. It can also strip other whitespace characters like tab `\t`

Assignment2/test/Assignment2.Tests/TestAddition.cs
```diff
@@ -0,0 +1,70 @@
...
+
+            string expectedOuput = $@">>Please enter a calculation to perform: <<{minVal}+{five}
+>>{minVal}+{five} = {res}
+>>Press Enter to Exit Program . . .";
+
+            IntelliTect.TestTools.Console.ConsoleAssert.Expect(expectedOuput, Assignment2.Program.Main);
+        }
+
+        [TestMethod]
+        public void TestAddMaxAndNeg7()
```
Keboo: For unit test names it is ok to be long and verbose. So use "Negative" rather than "Neg"

Assignment2/test/Assignment2.Tests/TestAddition.cs
```diff
@@ -0,0 +1,70 @@
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace Assignment2.Tests
+{
+    [TestClass]
+    public class TestAddition
+    {
+        [TestMethod]
+        public void TestAdd8And2HardCoded()
```
Keboo: You can omit "HardCoded" from the unit test method name. 
In general you can also omit "Test" from the unit test name since the fact that it is a unit test implies that it will be testing something.

Assignment2/src/Assignment2/Program.cs
```diff
@@ -0,0 +1,134 @@
...
+                else
+                {
+                    Console.WriteLine("Cannot Divide by Zero - Please Try Again");
+                }
+            }
+
+
+
+
+            else if (input.Contains('-')) //subtraction - by checking for subtraction last it ensures that negatives work properly for all above calculations
```
Keboo: Good use of comment explaining the "why" behind the code. The comments above that merely call out the name of the operator do not add much to the code, but this one does.

Keboo:
Good work. Nice job splitting up the unit tests to keep them small and easy to read.

PR #28, opened 2018-10-04
Assignment2 submission

COsborn2:


Assignment2/ConsoleMathSolver.Tests/StringTests.cs
```diff
@@ -0,0 +1,52 @@
...
+    [TestClass]
+    public class StringTests
+    {
+        // ends with
+        [TestMethod]
+        [DataRow("someinsidething", "something", "inside", 4)]
+        [DataRow("therehing", "thing", "here", 1)]
+        [DataRow("setfirst", "first", "set", 0)]
+        [DataRow("lastset", "last", "set", 4)]
+        public void TestStringInsert(string expected, string toInsertInto, string insertThis, int atIndex)
```
Keboo: Great use of data driven tests

Assignment2/ConsoleMathSolver/Program.cs
```diff
@@ -0,0 +1,210 @@
...
+
+        private static double CalculateValue(string userInput)
+        {
+            if (string.IsNullOrEmpty(userInput))
+            {
+                throw new InvalidDataException("Input cannot be empty!");
+            }
+            
+            // remove spaces from input string
+            string userInputNoSpaces = Regex.Replace(userInput, "\\ +", "");
```
Keboo: Good use of regex, but for simple character replacements like this you can also use the [string.Replace method](https://source.dot.net/#System.Private.CoreLib/shared/System/String.Manipulation.cs,1060).

Assignment2/ConsoleMathSolver/Program.cs
```diff
@@ -0,0 +1,210 @@
...
+            {
+                Console.WriteLine(e);
+                throw;
+            }
+        }
+
+
+        private static double CalculateValue(string userInput)
+        {
+            if (string.IsNullOrEmpty(userInput))
```
Keboo: There is also a `string.IsNullOrWhitespace` method that is a bit more common. This will save you from needing to check for an empty string after removing the whitespace.

Assignment2/ConsoleMathSolver/Program.cs
```diff
@@ -0,0 +1,210 @@
...
+            if (splitOnOperator.Count == 0)
+            {
+                throw new InvalidDataException("Invalid input sequence! Input must match <integer><operator><integer>");
+            }
+
+            int validNumCount = 0;
+
+            foreach (var temp in splitOnOperator)
+            {
+                if (int.TryParse(temp, out int parsedVal))
```
Keboo: Good use of `int.TryParse` and inline out parameters.

Assignment2/ConsoleMathSolver/Program.cs
```diff
@@ -0,0 +1,210 @@
...
+                        if (operatorInput[i + 1] == '-')
+                        {
+                            return "--";
+                        }
+                        else
+                        {
+                            return "-";
+                        }
+                    }
+                    else if (new List<char> {'+', '-', '/', '*'}.Contains(operatorInput[i]))
```
Keboo: Given that this collection is fixed, and is executed in a loop, it is best to declare the list/array outside of the loop so that it is only allocated once.

Assignment2/ConsoleMathSolver.Tests/ConsoleMathSolverTests.cs
```diff
@@ -0,0 +1,326 @@
...
+            double expectedAnswer = 1060;
+            string expectedOutput =
+$@"Please enter expression in form <integer><operator><integer>: <<{testVal}
+>>Value is: {expectedAnswer}";
+
+            IntelliTect.TestTools.Console.ConsoleAssert.Expect(expectedOutput, ConsoleMathSolver.Program.Main);
+        }
+        
+        [TestMethod]
+        public void PositivePlusPositive()
```
Keboo: Good job testing all of the various cases. However, when doing tests that take in multiple inputs, it is usually best to not use the same value for both inputs.

Keboo:
Very nice work. Easy to read and follow.

PR #30, opened 2018-10-04
Assignment2

Cwhite534:
Calculator and unit tests for binomial expression solver. Supported operators: +,-,*,/,%,^

### DIFF
--- a/Assignment2/Assignment2-Calculator/Calculator.cs
+++ b/Assignment2/Assignment2-Calculator/Calculator.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Collections.Generic;
 using System.Diagnostics;
 using System.Text.RegularExpressions;
 
@@ -14,16 +15,14 @@ namespace Assignment2_Calculator
 
             try
             {
-                (long operandOne, long operandTwo, char operatorOne)
-                    expressionTuple = ParseIntoTuple(userInput);
+                (long operandOne, long operandTwo, char operatorOne)expressionTuple = ParseIntoTuple(userInput);
 
                 var solution = EvaluateExpression(expressionTuple);
-                Console.WriteLine(
-                    $"{expressionTuple.operandOne}{expressionTuple.operatorOne}{expressionTuple.operandTwo}={solution}");
+                Console.WriteLine($"{expressionTuple.operandOne}{expressionTuple.operatorOne}{expressionTuple.operandTwo}={solution}");
             }
             catch (Exception e)
             {
-                Console.WriteLine(e.GetType());
+                Console.WriteLine(e.GetType() + e.ToString());
             }
         }
 
@@ -46,13 +45,11 @@ namespace Assignment2_Calculator
         private static (long, long, char) ParseIntoTuple(string expressionStr)
         {
             (long operandOne, long operandTwo, char operatorOne) expressionTuple; 
-            var regex = new Regex(@"(?<=\d)[-+*/]{1}");        //gets operator not minus sign
-
-            var match = regex.Match(expressionStr);
+            var match = Regex.Match(expressionStr,@"(?<=\d)[-+*/]{1}");                        //gets operator not negative sign        
             char.TryParse(match.Value, out expressionTuple.operatorOne);
-            string[] splitExpression = expressionStr.Split(expressionTuple.operatorOne);
-            long.TryParse(splitExpression[0], out expressionTuple.operandOne);
-            long.TryParse(splitExpression[1], out expressionTuple.operandTwo);
+            
+            long.TryParse(expressionStr.Substring(0,match.Index), out expressionTuple.operandOne);
+            long.TryParse(expressionStr.Substring(match.Index+1), out expressionTuple.operandTwo);
             return expressionTuple;
         }
 
--- a/Assignment2/Assignment2-Calculator/Calculator.cs
+++ b/Assignment2/Assignment2-Calculator/Calculator.cs
@@ -4,21 +4,20 @@ using System.Text.RegularExpressions;
 
 namespace Assignment2_Calculator
 {
-    public class Calculator
+    public static class Calculator
     {  
         public static void Main(string[] args)
         {   
             var userInput = args.Length == 0 ? TakeUserInput() : args[0];
-            int max = int.MaxValue;
             while(!IsValidExpression(userInput))
                 userInput = TakeUserInput();
 
             try
             {
-                (int operandOne, int operandTwo, char operatorOne)
-                    expressionTuple = ParseExpressionIntoTuple(userInput);
+                (long operandOne, long operandTwo, char operatorOne)
+                    expressionTuple = ParseIntoTuple(userInput);
 
-                double solution = EvaluateExpression(expressionTuple);
+                var solution = EvaluateExpression(expressionTuple);
                 Console.WriteLine(
                     $"{expressionTuple.operandOne}{expressionTuple.operatorOne}{expressionTuple.operandTwo}={solution}");
             }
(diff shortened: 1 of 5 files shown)


Assignment2/Assignment2-Calculator/Calculator.cs
```diff
@@ -0,0 +1,82 @@
...
+        }
+
+
+        private static bool IsValidExpression(string expressionStr)
+        {
+            var regex = new Regex(@"^-?\d+[-+/*%^]{1}-?\d+$");    //correct expression pattern
+            var isValid = regex.IsMatch(expressionStr);
+            regex = new Regex(@"-?\d+[/]{1}0+$");                //checks if there is division by zero
+            var hasDivisionByZero = regex.IsMatch(expressionStr);
+            regex = new Regex(@"-?\d+[%]{1}0+$");                //checks if there is division by zero
```
Keboo: Looks like a copy/paste error in the comment here.

Assignment2/Assignment2-Calculator/Calculator.cs
```diff
@@ -0,0 +1,82 @@
...
+        private static string TakeUserInput()
+        {
+            Console.Write("Please enter an expression ie. (3+1): ");
+            return Regex.Replace(Console.ReadLine(), " ", "");
+        }
+
+
+        private static bool IsValidExpression(string expressionStr)
+        {
+            var regex = new Regex(@"^-?\d+[-+/*%^]{1}-?\d+$");    //correct expression pattern
```
Keboo: Good use of regex. You could simplify it slightly since the repetition `{1}` is redundant. 
```C#
new Regex(@"^-?\d+[-+/*%^]-?\d+$")
```

Assignment2/Assignment2-Calculator/Calculator.cs
```diff
@@ -0,0 +1,82 @@
...
+            var isValid = regex.IsMatch(expressionStr);
+            regex = new Regex(@"-?\d+[/]{1}0+$");                //checks if there is division by zero
```
Keboo: With regex, if you have a set of items that only contains one item, you can remove the staples (`[` and `]`).
```C#
regex = new Regex(@"-?\d+/0+$");  
```

Assignment2/Assignment2-Calculator/Calculator.cs
```diff
@@ -0,0 +1,82 @@
...
+                Console.WriteLine($"Invalid Expression: {expressionStr}");
+            if(hasDivisionByZero)
+                Console.WriteLine($"Cannot divide by zero: {expressionStr}");
+            if(hasModByZero)
+                Console.WriteLine($"Cannot mod by zero: {expressionStr}");
+            
+            return isValid && !hasDivisionByZero && !hasModByZero;
+        }
+
+        private static (long, long, char) ParseIntoTuple(string expressionStr)
```
Keboo: Good use of `ValueTuple` but I would give each of the items a name to make it more obvious what is being returned.

Assignment2/Assignment2-Calculator/Calculator.cs
```diff
@@ -0,0 +1,82 @@
...
+        {
+            (long operandOne, long operandTwo, char operatorOne) expressionTuple;
```
Keboo: Minor naming thing here but it would be better to call the third element "operator" rather than "operatorOne" since there is only one.

Assignment2/CalculatorTest/CalculatorTester.cs
```diff
@@ -0,0 +1,560 @@
...
+         * Note that for tester simplicity i did not include spaces in the
+         * input with the exception of this method
+         */
+        [DataRow("5 + 5", "10")]
+        [DataRow("5 +9", "14")]
+        [DataRow("1+3", "4")]
+        [DataTestMethod]
+        public void UserInputPromptValid(string inputStr, string answer)
+        {
+            var inputStrNoSpace = Regex.Replace(inputStr, " ", "");
```
Keboo: Though Regex works, for simple character replacement `string.Replace` is a bit better.

Assignment2/CalculatorTest/CalculatorTester.cs
```diff
@@ -0,0 +1,560 @@
...
+    {
+        /**
+         * Note that for tester simplicity i did not include spaces in the
+         * input with the exception of this method
+         */
+        [DataRow("5 + 5", "10")]
+        [DataRow("5 +9", "14")]
+        [DataRow("1+3", "4")]
+        [DataTestMethod]
+        public void UserInputPromptValid(string inputStr, string answer)
```
Keboo: Good use of data driven tests!

Assignment2/CalculatorTest/CalculatorTester.cs
```diff
@@ -0,0 +1,560 @@
...
+        {
+            var ExpectedOutput = $@">>Please enter an expression ie. (3+1): <<{inputStr}
+            >>{inputStr}={answer}";
+            ConsoleAssert.Expect(ExpectedOutput, Calculator.Main);
+        }
+    }
+
+
+    [TestClass]
+    public class SubtractionIntegerTests
```
Keboo: Good job separating the tests into different classes. Typically we would place each class in its own corresponding file.

Assignment2/CalculatorTest/CalculatorTester.cs
```diff
@@ -0,0 +1,560 @@
...
+            ConsoleAssert.Expect(ExpectedOutput, Calculator.Main);
+        }
+
+        [DataRow("-50/5", "-10")]
+        [DataRow("-10000/20", "-500")]
+        [DataRow("-9999/9", "-1111")]
+        [DataRow("-100/10", "-10")]
+        [DataRow("-1/2", "-0.5")]
+        [DataTestMethod]
+        public void DivideNegDividePos(string inputStr, string answer)
```
Keboo: For unit tests it is ok to have long method names and be verbose. So avoid abbreviations and write everything out.

Assignment2/CalculatorTest/CalculatorTester.cs
```diff
@@ -0,0 +1,560 @@
...
+    [TestClass]
+    public class StringTests
+    {
+        [DataRow("Casey", "White")]
+        [DataRow("Easy", "Test")]
+        [DataRow("Cake", "Walk")]
+        [DataTestMethod]
+        public void TestLengthEqual(string str1, string str2)
+        {
+            Assert.AreEqual(str1.Length, str2.Length);
```
Keboo: This works, but typically for the first ("expected") parameter in the `Assert.AreEqual` you want it to be a constant expression rather than something that is calculated.

Assignment2/CalculatorTest/CalculatorTester.cs
```diff
@@ -0,0 +1,560 @@
...
+
+        [DataRow("Casey", "White")]
+        [DataRow("Easy", "Test")]
+        [DataRow("Cake", "Walk")]
+        [DataTestMethod]
+        public void TestStringCompareUnequal(string str1, string str2)
+        {
+            var comparison = str1.CompareTo(str2);
+            Assert.AreNotEqual(0, comparison);
+        }
```
Keboo: The compare method returns a negative number, or a positive number that can be used for sorting. A more complete test would have looked at both of those cases directly.

Keboo:
Very nice work.

Assignment2/Assignment2-Calculator/Calculator.cs
Keboo: Correct, but you can use keywords and variables names if you prefix them with an @ symbol. See chapter 1 (page 9).

PR #31, opened 2018-10-05
Assignment2

Chester-Southwood:


Assignment2/src/Expression.cs
```diff
@@ -0,0 +1,289 @@
+using System;
+using System.Collections.Generic;
+namespace Expression
+{
+    public class Expression
+    {
+    private String inFix;
```
Keboo: Favor using C# keywords over types names. So `string` instead of `String`

Assignment2/src/Expression.cs
```diff
@@ -0,0 +1,289 @@
...
+    private LinkedList<String> postFixLL;
+    private Stack expressionStack;
```
Keboo: Wow. The assignment only asked you to handle a single expression, not a nested stack of them. This is very good.

Assignment2/src/Expression.cs
```diff
@@ -0,0 +1,289 @@
...
+                {               
+                    if(!Decimal.TryParse(Math.Pow((double)right, (double)left).ToString(), out tempValue) || !checkValidDecimalValue(tempValue)) 
+                    {
+                        return "Error: Overflow would occur if attempt to take power to.";
+                    }    
+                    this.expressionStack.push(tempValue.ToString());
+                }
+                else if(string.Equals(symbol, "/"))
+                {
+                    if(left == 0)
```
Keboo: This seems like your variable names are backwards. The number on the "left" should be the numerator, not the denominator. Pretty sure this is just a naming issue as the program appears to work just fine.

Assignment2/src/Expression.cs
```diff
@@ -0,0 +1,289 @@
...
+        }
+        decimal finalDec =  Decimal.Round(Decimal.Parse(this.expressionStack.pop()),9);
+        if(Decimal.Equals(finalDec, 0))
+        {
+            finalDec = 0;
+        }
+        return this.inFix + " = " + finalDec;
+    }
+
+    private bool checkValidDecimalValue(decimal tempValue)
```
Keboo: I am not sure what the intent of this method is. Most decimals will be convertible to doubles since decimal has a smaller range of possible values. The only difference is that it will be accurate with its decimal places, while double will not. The check after against the decimal min/max value is also redundant since the incoming parameter was already a valid decimal.
The name of the method is also a bit confusing as it overloads the method below, but has a very different criteria.

Assignment2/src/Expression.cs
```diff
@@ -0,0 +1,289 @@
...
+        }
+        return false;
+    }
+
+    private bool checkValidDecimalValue(String tempValue)
+    {
+        decimal tempDoub = 0;
+        if(Decimal.TryParse(tempValue.ToString(), out tempDoub))
+        {
+            if(tempDoub > Decimal.MaxValue || tempDoub < Decimal.MinValue)
```
Keboo: This check is redundant. If `decimal.TryParse` returns true, then the resulting decimal in `tempDoub` will be in the valid range.

Assignment2/src/Expression.cs
```diff
@@ -0,0 +1,289 @@
...
+                return false;
+            }
+            return true;
+        }
+        return false;
+    }
+
+    private bool checkValidDecimalValue(String tempValue)
+    {
+        decimal tempDoub = 0;
```
Keboo: Variable name is confusing since it doesn't match its type. A better name would simply be `temp`

Assignment2/src/Stack.cs
```diff
@@ -0,0 +1,68 @@
...
+      else
+      {
+         return this.top.getData();
+      }
+   }
+   public String pop()
+   {
+      if(isEmpty())
+      {
+         return "No Expression";
```
Keboo: Given that the `peek` method throws an exception. I would expect similar behavior from `pop()`

Assignment2/tests/UnitTest1.cs
```diff
@@ -0,0 +1,2013 @@
+using Microsoft.VisualStudio.TestTools.UnitTesting;
```
Keboo: Next time rename your unit test file to match the test class inside of it. You never want to leave the file name as "UnitTest1.cs"

Assignment2/tests/UnitTest1.cs
```diff
@@ -0,0 +1,2013 @@
...
+            Assert.AreEqual("", "can".Substring(0,0));
+            Assert.AreEqual("bob", "bob".Substring(0,"bob".Length));
+            Assert.AreEqual("", "".Substring(0,0));
+            Assert.AreEqual("a", "andy".Substring(0,1));
+        }
+    }
+
+//CHECK MAX VALUE
+    [TestClass]
+    public class UnitTestEdgeCasesMax
```
Keboo: Next time separate each unit test class into its own file.

Assignment2/tests/UnitTest1.cs
```diff
@@ -0,0 +1,2013 @@
...
+    {
+        [TestMethod]
+        public void testDecMaxInputed()
```
Keboo: For unit test methods, it is ok to have long names and be verbose. Avoid abbreviations and spell everything out.

Assignment2/tests/UnitTest1.cs
```diff
@@ -0,0 +1,2013 @@
...
+>>Error: Overflow would occur if attempt to take power to."; 
+            IntelliTect.TestTools.Console.ConsoleAssert.Expect(expectedOutput,  src.Program.Main);
+        }
+    }
+
+
+
+//CHECK MIN VALUE
+    [TestClass]
+    public class UnitTestEdgeCasesMin
```
Keboo: Typically you don't need to prefix unit test classes with "UnitTest". The fact that they have the `[TestClass]` attribute is sufficient to know that they contain unit tests.

Assignment2/src/Expression.cs
```diff
@@ -0,0 +1,289 @@
...
+    private void postFixInit()
+    {
+        this.postFixLL = new LinkedList<string>();
+        int index = 0;
+        bool invalidSym = false;
+        while(index < inFix.Length)
+        {  
+            char symbol = inFix[index];
+            //                                 check for decimal value         
+            if((symbol > 47 && symbol < 58) || symbol == '.' || (inFix.Length > 2 &&  symbol == '-'  && inFix[index + 1] == '.' && (inFix[index + 2] > 47 && inFix[index + 1] < 58) ) || (symbol == '-' && (inFix[index + 1] > 47 && inFix[index + 1] < 58)))
```
Keboo: Avoid doing character comparison by their ASCII (int) values. A better test here would have been to use the `char.IsDigit` method.

Assignment2/src/Expression.cs
```diff
@@ -0,0 +1,289 @@
...
+                {
+                    return "Error: Invalid Expression cannot be calculated.";
+                }
+            }
+            index++;
+        }
+        decimal finalDec =  Decimal.Round(Decimal.Parse(this.expressionStack.pop()),9);
+        if(Decimal.Equals(finalDec, 0))
+        {
+            finalDec = 0;
```
Keboo: This setter appears redundant. If `finalDec` is equal to 0, then setting it to 0 does nothing.

Keboo:
Good work. I was impressed you handled nested expressions. It appears you ended up including some files from assignment 1 as well as a bunch of files that should have been ignored. For assignment 3 make sure you are sync'd up with the latest code on IntelliTect/master and this should not be a problem.

PR #37, opened 2018-10-05
Assignment2

valdynhunt:
Submit assignment2.


Assignment2/MathExpressions/src/MathExpressions/MathExpressions.cs
```diff
@@ -0,0 +1,241 @@
...
+        }
+
+        public void EvaluateExpression()
+        {
+            firstOperand = parts[0];
+            lastOperand = parts[1];
+
+            bool divZero = ((op == '/') && CheckDivideByZero(lastOperand));
+            bool firstInRange = ValidateFirstOperandInRange(firstOperand);
+            bool lastInRange = ValidateLastOperandInRange(lastOperand);
```
Keboo: `firstInRange` and `lastInRange` are never used.

Assignment2/MathExpressions/test/MathExpressions.Tests/MathExpressionsTests.cs
```diff
@@ -0,0 +1,95 @@
...
+            Assert.AreEqual(expectedOutput, MathExpressions.FindOperator(expr));
+
+            expr = "8/4";
+            expectedOutput = '/';
+            Assert.AreEqual(expectedOutput, MathExpressions.FindOperator(expr));
+
+            expr = "6-5";
+            expectedOutput = '-';
+            Assert.AreEqual(expectedOutput, MathExpressions.FindOperator(expr));
+        }
```
Keboo: There is an additional case that `FindOperator` can return `\0` that is not tested. 
Typically you would separate each of these cases into their own unit tests.

Assignment2/MathExpressions/src/MathExpressions/MathExpressions.cs
```diff
@@ -0,0 +1,241 @@
...
+                good = ValidateExpr(expr);
+            }
+        }
+
+        public void SplitString()
+        {
+            parts = this.expr.Split(this.op);
+        }
+
+        public bool ValidateOperands()
```
Keboo: This method always returns `false` and its return value does not appear to be used.

Assignment2/MathExpressions/src/MathExpressions/MathExpressions.cs
```diff
@@ -0,0 +1,241 @@
...
+        public bool ValidateExpr(string expr)
+        {
+            bool good = false;
+            Console.WriteLine("You entered {0}", this.expr);
+            this.op = FindOperator(this.expr);
+
+            if (this.op == '\0')
+            {
+                Console.WriteLine("Your entry is invalid, please try again...");
+                this.expr = Console.ReadLine();
```
Keboo: This attempts to re-read a new expression but this new expression does not get re-validated.

PR #43, opened 2018-10-11
Assignment3

Cwhite534:


Assignment3/RockPaperScissors/RockPaperScissors.cs
```diff
@@ -0,0 +1,119 @@
+﻿using System;
+
+namespace RockPaperScissors
+{
+    public class RockPaperScissors
+    {
+        public static void Main(string[] args)
```
breaman: I like how clean and concise your main method is. Nice work moving as much logic out as possible.

Assignment3/RockPaperScissors/RockPaperScissors.cs
```diff
@@ -0,0 +1,119 @@
...
+            
+            if (userInput.Equals("rock") || userInput.Equals("r"))
+                return "rock";
+            if (userInput.Equals("paper") || userInput.Equals("p"))
+                return "paper";
+            if(userInput.Equals("scissors") || userInput.Equals("s"))
+                return "scissors";
+         
+            Console.WriteLine("Invalid input");
+            return GetUserMove();
```
breaman: Nice use of recursion here, very clever.

Assignment3/RockPaperScissors/RockPaperScissors.cs
```diff
@@ -0,0 +1,119 @@
...
+            return playersHealth;
+        }
+        
+        
+        //Calculates winner and returns damage dealt. If positive, the user won, if negative, the computer won
+        public static int CalculateDamageDealt((string userMove, string cpuMove) roundMoves)
+        {
+            if (roundMoves.userMove.Equals(roundMoves.cpuMove))
+                return 0;
+            switch (roundMoves.userMove)
```
breaman: This was very well done on how you calculated the winner/loser and the point damage. I had a bunch of if statements, but this is really slick.

Assignment3/RockPaperScissorsTest/TestCalculateDamageDealt.cs
```diff
@@ -0,0 +1,78 @@
...
+            Assert.AreEqual(10, RockPaperScissors.RockPaperScissors.CalculateDamageDealt(("paper", "rock")));
+        }
+        
+        [TestMethod]
+        public void TestUserBeatsCpuWithScissors()
+        {
+            Assert.AreEqual(15, RockPaperScissors.RockPaperScissors.CalculateDamageDealt(("scissors", "paper")));
+        }
+        
+        //cpu wins
```
breaman: I like the explicitness, so I'm torn writing this comment, but I wonder if using the data driven tests for this would still have been better, and then just having tests of "UserWins", "CpuWins", and "Draw" with the 3 conditions as DataRow(point, userMove, cpuMove).

PR #44, opened 2018-10-11
Assignment 3

joeriddles:
CSCD371 Assignment 3 submission.

### DIFF
--- a/Assignment3/Roshambo/Roshambo/Roshambo.cs
+++ b/Assignment3/Roshambo/Roshambo/Roshambo.cs
@@ -29,7 +29,7 @@ namespace Roshambo
 
 				CalculateWinner(playerMove, computerMove);
 
-				PrintHealth();
+				Console.WriteLine($"Player Health: {playerHealth}, Computer Health: {computerHealth}");
 			}
 
 			if (playerHealth <= 0 && computerHealth <= 0)
@@ -77,11 +77,6 @@ namespace Roshambo
 			}
 
 			throw new ArgumentException();
-		}
-
-		private static void PrintHealth()
-		{
-			Console.WriteLine($"Player Health: {playerHealth}, Computer Health: {computerHealth}");
 		}
 
 		public static (int winner, int damage) CalculateWinner(RoshamboMove playerMove, RoshamboMove computerMove)
--- a/Assignment3/Roshambo/Roshambo/Roshambo.cs
+++ b/Assignment3/Roshambo/Roshambo/Roshambo.cs
@@ -87,33 +87,33 @@ namespace Roshambo
 					switch (playerMove)
 					{
 						case RoshamboMove.paper:
-							computerHealth -= paperDamage;
-							return (-1, paperDamage);
+							computerHealth -= (int)RoshamboMove.paper;
+							return (-1, (int)RoshamboMove.paper);
 						case RoshamboMove.scissors:
-							playerHealth -= rockDamage;
-							return (1, rockDamage);
+							playerHealth -= (int)RoshamboMove.rock;
+							return (1, (int)RoshamboMove.rock);
 						default: return (0, 0);
 					}
 				case RoshamboMove.paper:
 					switch (playerMove)
 					{
 						case RoshamboMove.rock:
-							playerHealth -= paperDamage;
-							return (1, paperDamage);
+							playerHealth -= (int)RoshamboMove.paper;
+							return (1, (int)RoshamboMove.paper);
 						case RoshamboMove.scissors:
-							computerHealth -= scissorsDamage;
-							return (-1, scissorsDamage);
+							computerHealth -= (int)RoshamboMove.scissors;
+							return (-1, (int)RoshamboMove.scissors);
 						default: return (0, 0);
 					}
 				case RoshamboMove.scissors:
 					switch (playerMove)
 					{
 						case RoshamboMove.rock:
-							computerHealth -= rockDamage;
-							return (-1, rockDamage);
+							computerHealth -= (int)RoshamboMove.rock;
+							return (-1, (int)RoshamboMove.rock);
 						case RoshamboMove.paper:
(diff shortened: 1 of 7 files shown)


Assignment3/Roshambo/Roshambo/Roshambo.cs
```diff
@@ -0,0 +1,132 @@
...
+using System.Linq;
+
+namespace Roshambo
+{
+	public class Roshambo
+	{
+		private static void Main(string[] args)
+		{
+			string playAgain = "y";
+			while (playAgain.Equals("y"))
```
breaman: This would have been a good place to use a do/while instead of a while since you knew you wanted to run through the loop at least once.

Assignment3/Roshambo/Roshambo/Roshambo.cs
```diff
@@ -0,0 +1,132 @@
...
+			{
+				LaunchGame();
+				Console.WriteLine("Play again? (y/n)");
+				playAgain = Console.ReadLine().Trim();
```
breaman: nice work on using the trim() and I know I said you wouldn't have to worry about casing, but would be good practice to call ToLower() here as well. Don't worry, not marking down, just leaving a comment since you did more than the basic of just reading the line.

Assignment3/Roshambo/Roshambo/Roshambo.cs
```diff
@@ -0,0 +1,132 @@
...
+			}
+		}
+
+		public static (int playerHealth, int computerHealth) LaunchGame()
```
breaman: This method returns a tuple, but you never do anything with the values. What was the reasoning behind this?

Assignment3/Roshambo/Roshambo/Roshambo.cs
```diff
@@ -0,0 +1,132 @@
...
+		{
+			string[] validInput = new string[] { "rock", "r", "scissors", "s", "paper", "p" };
+
+			while (!validInput.Contains(playerMove))
+			{
+				Console.WriteLine("'Rock', 'Scissors', or 'Paper'?");
+				playerMove = Console.ReadLine().Trim();
+			}
+
+			switch (playerMove)
```
breaman: here is where you should have done something like
```
switch (playerMove)
{
   case "rock":
   case "r": return RoshamboMove.rock;
   ...
}
```

Assignment3/Roshambo/Roshambo/Roshambo.cs
```diff
@@ -0,0 +1,132 @@
...
+						case RoshamboMove.paper:
+							playerHealth -= (int)RoshamboMove.scissors;
+							return (1, (int)RoshamboMove.scissors);
+						default: return (0, 0);
+					}
+				default: throw new ArgumentException();
+			}
+		}
+
+		public enum RoshamboMove
```
breaman: Nice use of the enum to store the move and the point value.

Assignment3/Roshambo/RoshamboTest/RoshamboTest/RoshamboTestGetComputerMove.cs
```diff
@@ -0,0 +1,19 @@
...
+	[TestClass]
+	public class RoshamboTestGetComputerMove
+	{
+		[TestMethod]
+		public void TestGetComputerMoves()
+		{
+			Roshambo.Roshambo.RoshamboMove[] roshamboMoves = new[] {Roshambo.Roshambo.RoshamboMove.paper, Roshambo.Roshambo.RoshamboMove.rock , Roshambo.Roshambo.RoshamboMove.scissors };
+
+			for (int i = 0; i < 100; i++)
+				Assert.IsTrue(roshamboMoves.Contains(Roshambo.Roshambo.GetComputerMove()));
```
breaman: What is this actually validating? Not sure what the loop to 100 is supposed to be accomplishing.

Assignment3/Roshambo/Roshambo/Roshambo.cs
```diff
@@ -0,0 +1,132 @@
...
+				RoshamboMove computerMove = GetComputerMove();
+
+				Console.WriteLine($"Computer used {computerMove}!");
+
+				CalculateWinner(playerMove, computerMove);
+
+				Console.WriteLine($"Player Health: {playerHealth}, Computer Health: {computerHealth}");
+			}
+
+			if (playerHealth <= 0 && computerHealth <= 0)
```
breaman: Can this condition ever actually happen? Since both can't "lose" at the same time, I don't believe this if check will ever be true.

Assignment3/Roshambo/Roshambo/Roshambo.cs
joeriddles: Good point. I forget to use do/whiles.

Assignment3/Roshambo/Roshambo/Roshambo.cs
joeriddles: Great point, thank you for pointing that out.

Assignment3/Roshambo/RoshamboTest/RoshamboTest/RoshamboTestGetComputerMove.cs
joeriddles: The goal here was to show that the computer will **always** generate a valid move.

Assignment3/Roshambo/Roshambo/Roshambo.cs
joeriddles: The reasoning behind this was purely because the assignment required a method that returned a tuple.

Assignment3/Roshambo/Roshambo/Roshambo.cs
joeriddles: That's very true: this condition would never be true. I'll remove it.

PR #46, opened 2018-10-12
Assignment3

JeffHowes:
Pull request for assignment 3


Assignment3/HW3-Roshambo/Player.cs
```diff
@@ -0,0 +1,103 @@
...
+        private int playersPoints;
+        private string playersWeaponChoice;
+        private string playersName;       
+        
+        public Player(string name)
+        {            
+            playersName = name;
+        }
+
+        public void initialPlayersPoint()
```
breaman: You have some camelCase methods and some PascalCase methods. C# uses PascalCase and you will be marked down in the future if you don't use PascalCase for method names. Either way, you should have picked one fashion and stuck with it throughout.

Assignment3/HW3-Roshambo/Player.cs
```diff
@@ -0,0 +1,103 @@
...
+            else                
+                playersWeaponChoice = "scissors";
+        }
+
+        public string getWeaponPicked()
+        {
+            return playersWeaponChoice;
+        }
+
+        public void SubtractPoints()
```
breaman: There is a logic flaw here. When playing the game, I got this behavior (it was the first move of the game, so we both had 100 to start out with.
```
You picked rock
The computer picked paper
The computer won this round!
You have 80 points and the computer has 100 points
```

Assignment3/RoshamboUnitTest/UnitTest1.cs
```diff
@@ -0,0 +1,266 @@
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using System;
+using System.IO;
+
+namespace RoshamboUnitTest
+{
+    [TestClass]
+    public class UnitTest1
```
breaman: This class should be better named than just UnitTest1.

Assignment3/RoshamboUnitTest/UnitTest1.cs
```diff
@@ -0,0 +1,266 @@
...
+            // Act            
+            human.validHumansWeaponChoice("Lizzard");
+
+            // Assert
+            Assert.AreEqual("rock", human.getWeaponPicked());
+            
+        }//end of test method
+
+        [TestMethod]
+        public void SubtractPoints_ForRock_100PointsDownTo80()
```
breaman: What are the SubtractPoints tests actually testing? You are setting the weapon and then subtracting points from the same player who has the weapon? Shouldn't it subtract 20 from computer if the human won and the human's weapon of choice was a rock?

Assignment3/RoshamboUnitTest/UnitTest1.cs
breaman: You are missing tests for your AdjustPoints method which would have caught your logic flaw on points being subtracting correctly based on what object the winner used, not what object the loser used.

Assignment3/HW3-Roshambo/Player.cs
JeffHowes: Mike, I don't see the logic error here.  Paper bets rock, so the computer should have won as it did and the users points should have been reduced, as they were by 20 points (per the requirements). Can you please explain the bug you are seeing?

PR #47, opened 2018-10-12
Raymond Shiner - Assignment 3 Submission

raymondshiner:
Did not unit test main because there's no real way to do it with variable input and do/while loop

All of the methods (5 of them) are tested fully.

I did not test for edge cases such as null strings being input by user based on the README.md assuring specific input strings at certain points in the program. (For example - my unit test on user inputing what theyd like to throw only handles "rock", "paper", and all other cases are assigned to a value of "scissors" - etc.)

### DIFF
--- a/Assignment3/src/Assignment3/Program.cs
+++ b/Assignment3/src/Assignment3/Program.cs
@@ -1,0 +1,46 @@
+﻿using System;
+
+namespace Assignment3
+{
+    class Program
+    {
+        static void Main(string[] args)
+        {
+            bool goAgain;
+
+            do
+            {
+                Console.WriteLine("Welcome to a game of Rock Paper Scissors");
+                Console.WriteLine("Both Players start with 100 life, its you versus the Computer");
+                Console.WriteLine("Good Luck - Have Fun");
+                Console.WriteLine();
+
+                int cpuLife = 100;
+                int userLife = 100;
+
+                while (true)
+                {
+                    if (cpuLife <= 0 || userLife <= 0)
+                    {
+                        break;
+                    }
+
+                    RoshamboMethods.PrintStatus(cpuLife, userLife);
+
+                    (string cpuThrow, string userThrow) throws = RoshamboMethods.GetThrows();
+
+                    RoshamboMethods.PrintThrowWinner(ref cpuLife, ref userLife, throws);
+                }
+
+                RoshamboMethods.PrintFinalScores(cpuLife, userLife);
+
+                goAgain = RoshamboMethods.GoAgain();
+
+            } while (goAgain);
+
+            Console.WriteLine("Thanks for Playing - Have a nice day");
+            Console.Write("Press any key to exit. . .");
+            Console.ReadLine();
+        }
+    }
+}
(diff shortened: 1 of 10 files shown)


Assignment3/src/Assignment3/Program.cs
```diff
@@ -0,0 +1,46 @@
...
+                    (string cpuThrow, string userThrow) throws = RoshamboMethods.GetThrows();
+
+                    RoshamboMethods.PrintThrowWinner(ref cpuLife, ref userLife, throws);
+                }
+
+                RoshamboMethods.PrintFinalScores(cpuLife, userLife);
+
+                goAgain = RoshamboMethods.GoAgain();
+
+            } while (goAgain);
```
breaman: I would eliminate the goAgain variable and just have a while(RoshamboMethods.GoAgain());

Assignment3/src/Assignment3/Program.cs
```diff
@@ -0,0 +1,46 @@
...
+            {
+                Console.WriteLine("Welcome to a game of Rock Paper Scissors");
+                Console.WriteLine("Both Players start with 100 life, its you versus the Computer");
+                Console.WriteLine("Good Luck - Have Fun");
+                Console.WriteLine();
+
+                int cpuLife = 100;
+                int userLife = 100;
+
+                while (true)
```
breaman: why not do/while this one as well? seems like we want to do things until cpuLife or userLife are <= 0.

Assignment3/test/Assignment3.Tests/TestPrintThrowWinner.cs
```diff
@@ -0,0 +1,350 @@
...
+
+            Assert.AreEqual(expectedOutput, programOutput);
+        }
+
+        //For Below Tests - 
+        // Win with Rock - Opponent Loses 20 life points
+        // Win with Paper - Opponent Loses 15 life points
+        // Wind with Scissors - Opponent Loses 10 life points
+
+        [TestMethod]
```
breaman: Nice job with the unit testing. Might want to look into using DataRow with the TestMethod in the future when you have multiple inputs testing the same method. Will save you some typing as well as make the unit tests easier to modify in the future in case inputs or outputs change.

Assignment3/test/Assignment3.Tests/TestPrintThrowWinner.cs
```diff
@@ -0,0 +1,350 @@
...
+        public void LifeChanges_UserRockCpuPaper()
+        {
+            int cpuLife = 100;
+            int userLife = 100;
+            (string cpuThrow, string userThrow) throws = ("paper", "rock");
+
+            Assignment3.RoshamboMethods.PrintThrowWinner(ref cpuLife, ref userLife, throws);
+
+            Assert.AreEqual(cpuLife, 100);
+            Assert.AreEqual(userLife, 100-15);
```
breaman: I understand what you are doing here because you want to make sure that the proper value is subtracted. What might be better in the future is to make a constant at the top of the file for the point loss value just in case that ever changes. Then you just need to change your value in one spot instead of looking for all the tests that subtracted a specific number.

PR #48, opened 2018-10-12
Assignment3

kveselits:
Only remaining issue I saw was that I don't have a green bar beside most of my CalculateHealth() method in Visual Studio. This seems odd since about 9 of my unit tests run on this method.


Assignment3/RockPaperScissors/Program.cs
```diff
@@ -0,0 +1,169 @@
...
+                return "rock";
+            }
+            else if (randomNumber.Equals(1))
+            {
+                return "paper";
+            }
+            return "scissors";
+        }
+
+        public static string DetermineWinner((int, int) health)
```
breaman: This would have been a little easier to read (and more self documenting) if you would have named your tuple values
```
public static string DetermineWinner((int player, int cpu) health)
        {
            if (health.cpu <= 0 && health.player > 0)
            {
                return ($"You win! {health.ToString()}");
            }
```

Assignment3/RockPaperScissors/Program.cs
```diff
@@ -0,0 +1,169 @@
...
+            }
+            else if ((health.Item2 > 0 && health.Item1 > 0))
+            {
+                return ($"No winner yet. {health.ToString()}");
+            }
+            return ($"Sorry, you lost. {health.ToString()}");
+        }
+
+
+        public static ValueTuple<int, int> CalculateHealth((int, int) health, string playerMove, string computerMove)
```
breaman: You don't need to use ValueTuple<int, int>, you can use the shorthand notation of just (int, int).

Assignment3/RockPaperScissorsUnitTests/RockPaperScissorsUnitTests.cs
```diff
@@ -0,0 +1,110 @@
...
+
+namespace RockPaperScissors
+{
+    [TestClass]
+    public class RockPaperScissorsUnitTests
+    {
+        ValueTuple<int, int> tempTuple = ValueTuple.Create(100, 100);
+
+        [TestMethod]
+        public void TestCalculateDamageRockRock()
```
breaman: Take a look at the [DataRow] attribute for data driven tests when doing something like this in the future.

Assignment3/RockPaperScissorsUnitTests/RockPaperScissorsUnitTests.cs
```diff
@@ -0,0 +1,110 @@
...
+        {
+            ValueTuple<int, int> health = ValueTuple.Create(100, 0);
+            string whoWins = Program.DetermineWinner(health);
+            Assert.AreEqual($"You win! {health.ToString()}", whoWins);
+
+        }
+        [TestMethod]
+        public void TestDetermineWinnerComputerWins()
+        {
+            ValueTuple<int, int> health = ValueTuple.Create(0, 100);
```
breaman: This could have just been (int, int) health = (0, 100); Not marking down, just showing a more compact syntax.

PR #49, opened 2018-10-12
Assignment3

NikolasGraham:
I was struggling with figuring out how to use unit testing still (I thought I could figure it out, but this class is my first time ever using them), and I failed to understand how to connect them and structure your code from your unit tests to be able to allow them to be connected, and test what needs to be tested. After today's class I feel better about understanding how to do them, but couldn't figure out how to put them into my code, and didn't believe I had enough time to redo my project from scratch. I apologize, and will work harder now that I have a better understanding for future assignments.

### DIFF
--- a/Assignment3/Roshambo/Roshambo/Roshambo.cs
+++ b/Assignment3/Roshambo/Roshambo/Roshambo.cs
@@ -1,0 +1,127 @@
+﻿using System;
+
+namespace Roshambo
+{
+    public class Roshambo
+    {
+        public static void Main(string[] args)
+        {
+            Console.WriteLine("<<Welcome to Roshambo>>\n");
+            string endGame = "";
+
+            do
+            {
+                endGame = "";
+                PlayGame();
+
+                Console.Write("\nPlay Again? (y/n)\n" +
+                    "-->");
+
+                while (endGame != "n" && endGame != "N" && endGame != "y" && endGame != "Y")
+                {
+                    endGame = Console.ReadLine();
+                    if(endGame != "n" && endGame != "N" && endGame != "y" && endGame != "Y")
+                    {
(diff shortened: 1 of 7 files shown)


Assignment3/Roshambo/Roshambo/Roshambo.cs
```diff
@@ -0,0 +1,127 @@
...
+            while (userInput.ToLower() != "rock" && userInput.ToLower() != "paper" && userInput.ToLower() != "scissors")
+            {
+                Console.Write("\nInvalid Move, try again...(Rock/Paper/Scissors)\n" +
+                    "-->");
+                userInput = Console.ReadLine();
+            }
+
+            Console.WriteLine("\n");
+
+            return "userInput";
```
breaman: This line is not doing what you are expecting is my guess. This always returns the string "userInput" instead of "rock/paper/scissors" which makes the game unplayable unfortunately.

Assignment3/Roshambo/test/RoshamboTests/UnitTestsRoshambo.cs
```diff
@@ -0,0 +1,19 @@
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace RoshamboTests
+{
+    [TestClass]
+    public class UnitTestsRoshambo
+    {
+        [TestMethod]
+        public void TestUserInputRock()
```
breaman: Yeah, I know you commented on this in your PR, so not trying to harp on the subject, but the idea was not necessarily to test any console input this time around, but more to test your methods that had your game logic in it. With the way the requirements were written, you would not have been able to really use the ConsoleTester stuff since you don't have a finite end to the game (since the computer selects a value at random, we can't know ahead of time how many iterations it would go through to determine a winner).

PR #51, opened 2018-10-12
Assignment3 Submission

KennethWhite:


Assignment3/Assignment3/Assignment3.src/Game.cs
```diff
@@ -0,0 +1,173 @@
...
+                {
+                    Console.WriteLine("Invalid option, please retry.");
+                }
+
+            }
+            Console.WriteLine("\nGame complete.");
+        }
+
+
+        public struct Player
```
breaman: We haven't hit the chapter where this is discussed yet, but this should have been a class instead of a struct. Just pointing it out for future reference. Of course by making this a struct, you got to experience the concept of out parameters on your methods :)

Assignment3/Assignment3/Assignment3.src/Game.cs
```diff
@@ -0,0 +1,173 @@
...
+            int health;
+            string lastMove;
+
+            public Player(string n = "Player", int h = 100, string lm = "")
+            {
+                name = n;
+                health = h;
+                lastMove = lm;
+            }
+            public int Health { get => health; set => health = value; }
```
breaman: These could have just been auto-implemented properties to save you some typing
```
public int Health {get; set;}
```

Assignment3/Assignment3/Assignment3.src/RockPaperScissors.cs
```diff
@@ -0,0 +1,77 @@
...
+            case "scissors":
+                return SCISSORS.ordinal;
+            default:
+                throw new ArgumentException($"'{name}' is not a valid option!");
+            }
+        }
+
+        public static int Compare(int lhsOrdinal, int rhsOrdinal)
+        {
+            if (lhsOrdinal == rhsOrdinal)
```
breaman: Clever way of determining the winner. Never thought of this approach before.

Assignment3/Assignment3/Assignment3.src/RockPaperScissors.cs
```diff
@@ -0,0 +1,77 @@
...
+            case "paper":
+                return PAPER.ordinal;
+            case "scissors":
+                return SCISSORS.ordinal;
+            default:
+                throw new ArgumentException($"'{name}' is not a valid option!");
+            }
+        }
+
+        public static int Compare(int lhsOrdinal, int rhsOrdinal)
```
breaman: Just a curiosity question here, why make this method static, but the other methods are instance based? You needed to create an instance of the class anyway, so why not make this one instance based as well?

KennethWhite: I can't quite recreate my mindset for making this static instead of instance based. I think since I wrote the method to accept an Int ordinal, which relied upon calling the static GetOrdinalByName method, it made sense to me to make this a static method.

PR #53, opened 2018-10-12
Assignment3

blazinaj:


Assignment3/Assignment3/Game.cs
```diff
@@ -0,0 +1,109 @@
...
+            {
+                string humanWeapon = humanPlayer.DrawWeapon();
+                Console.WriteLine($"--------------{Environment.NewLine}You have drawn: {humanWeapon}");
+
+                string computerWeapon = computerPlayer.DrawWeapon();
+                Console.WriteLine($"The computer has drawn: {computerWeapon}{Environment.NewLine}{Environment.NewLine}");
+
+                var weaponSelection = MapWeapons(humanWeapon, computerWeapon);
+
+                if (weaponSelection.BattleResult("rock", "rock"))
```
breaman: This was where most of the logic existed for the application. Where were the unit tests around BattleResult?

Assignment3/Assignment3/Draw.cs
```diff
@@ -0,0 +1,59 @@
...
+                case "paper":
+                    return "paper";
+                case "scissors":
+                    return "scissors";
+                default:
+                    return SelectWeapon();
+            }
+        }
+
+        override
```
breaman: We haven't covered this in class yet, but this probably isn't the right way to override ToString since weapon will never get set if it is a human player. You want ToString to always return some value.

Assignment3/Assignment3/Game.cs
```diff
@@ -0,0 +1,109 @@
...
+                    else
+                    {
+                        winner = "You, the human player";
+                    }
+                }
+
+            } while (winner == "none");
+        }
+
+        private Tuple<string, string> MapWeapons(string humanWeapon, string computerWeapon)
```
breaman: this could have been written as
```
private (string, string) MapWeapons(string humanWeapon, string computerWeapon)
{
   return (humanWeapon, computerWeapon);
}
```

Assignment3/Assignment3/BattleWeapons.cs
```diff
@@ -0,0 +1,14 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Assignment3
+{
+    public static class BattleWeapons
+    {
+        public static bool BattleResult<Weapon1, Weapon2>(this Tuple<Weapon1, Weapon2> weapons, Weapon1 weapon1Type, Weapon2 weapon2Type)
```
breaman: nice to see the use of an extension method here :)

PR #54, opened 2018-10-12
Andrew Fleiger - Assignment 3 Submission

afleiger:


Assignment3/Roshambo/roshambo.tests/Roshambo.Tests.cs
```diff
@@ -0,0 +1,236 @@
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using IntelliTect.TestTools.Console;
+using System;
+using System.Text;
+using System.IO;
+
+namespace Roshambo.Tests
+{
+    [TestClass]
```
breaman: Nice work on the unit tests. You might want to look into using [DataRow] in the future so that you don't have to duplicate so many tests.

Assignment3/Roshambo/roshambo.tests/Roshambo.Tests.cs
```diff
@@ -0,0 +1,236 @@
...
+    public class RoshamboUnitTest
+    {
+        [TestMethod]
+        public void RandomChoice_ReturnsValid()
+        {
+            //arrange
+            //act
+            for (int x = 0; x < 100; x++)
```
breaman: My guess is you are just running through a bunch of iterations to make sure that only rock, paper, or scissors is chosen, but not sure what that is really validating. Might want to make sure that all 3 options are chosen as well. Technically, your RandomChoice could be hard coded to return "rock" and this test would still pass.

Assignment3/Roshambo/roshambo.tests/Roshambo.Tests.cs
```diff
@@ -0,0 +1,236 @@
...
+        {
+            ConsoleAssert.Expect(@">>
+Enter 'rock', 'paper', or 'scissors':<<paper", () => Roshambo.PlayRound());
+
+        }
+
+        [TestMethod]
+        public void PlayRound_SelectsPaper_ReturnsTupleWithPaperSelected()
+        {
+            string userInput = "paper\n";
```
breaman: Nice testing around user input.

Assignment3/Roshambo/src/Roshambo.cs
```diff
@@ -0,0 +1,123 @@
...
+                case 2:
+                    ret = "scissors";
+                    break;
+            }
+            return ret;
+        }
+
+        public static int DetermineDamage(string player, string opponent)
+        {
+            switch (player)
```
breaman: This was a great way of determining the winner. Just call it twice like you did, instead of a giant nested if statement.

PR #55, opened 2018-10-12
attemptToUpdate Assignment3

Chester-Southwood:
Attempted to branch properly and pull for Assignment3. I may've added previous commits by accident. Assignment3 should contain the sln file, src, and tester files. 
Tests were run by the logic based methods excluding the main method. No error checking for false input as the input is guaranteed in the directions.


Assignment3/tester/UnitTest1.cs
```diff
@@ -0,0 +1,387 @@
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using System;
+namespace tester
+{
+    [TestClass]
+    public class PlayRoundTest
```
breaman: Make sure your filename matches your class name. Makes it easier to find things.

Assignment3/src/Program.cs
```diff
@@ -0,0 +1,208 @@
+﻿using System;
+
+
+namespace Ass3
+{
+    public class Player
```
breaman: While not required by the language, it is usually considered best practice to have one class per file.

Assignment3/src/Program.cs
```diff
@@ -0,0 +1,208 @@
...
+                min = temp.returnMin;
+                max = temp.returnMax;
+                Console.Write("Do you wish to play again? -> ");
+                playAgain = Console.ReadKey().KeyChar;
+                Console.WriteLine(Environment.NewLine);
+            }while(playAgain == 'y');
+            Console.WriteLine(DeclareWinner(user, ai));
+        }
+
+        public (int returnMin, int returnMax) BestOutOf(int min, int max, Player user, Player ai)
```
breaman: Clever work on allowing the loser to always try and win things back win the "best 2 out of 3", "3 out of 5", etc.

Assignment3/tester/UnitTest1.cs
```diff
@@ -0,0 +1,387 @@
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using System;
+namespace tester
+{
+    [TestClass]
+    public class PlayRoundTest
+    {
+        [TestMethod]
+        public void TestRockVsPaper()
```
breaman: This group of tests would have been a perfect candidate for using the data driven testing and the DataRow attribute. Not marking down, just might want to consider it for the future.

PR #56, opened 2018-10-12
Assignment3

toha87:
Tried my best! :)


Assignment3/PaperRockScissors/GameUnitTest/GameUnitTest.cs
```diff
@@ -0,0 +1,72 @@
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using PaperRockScissors;
+using System;
+
+namespace GameUnitTest
+{
+    [TestClass]
```
breaman: You didn't have any unit tests around your GameAction method which is where most of your logic lived. Would have been good to break out the logic for determining the winner of each "round" so you could have tested that logic as well.

Assignment3/PaperRockScissors/PaperRockScissors/PaperRockScissorsClass.cs
```diff
@@ -0,0 +1,168 @@
+﻿using System;
+
+namespace PaperRockScissors
+{
+    public class PaperRockScissorsClass
+    {
+        public static void Main(string[] args)
+        {
+            bool PlayAgain = true;
```
breaman: local variables should be camelCase instead of PascalCase

Assignment3/PaperRockScissors/PaperRockScissors/PaperRockScissorsClass.cs
```diff
@@ -0,0 +1,168 @@
...
+                        PlayAgain = PlayAgainAnswer;
+
+                } while (PlayAgain == true);
+        }
+
+        public static string GetUserInput()
+        {
+
+                Console.WriteLine("Do you choose rock, paper or scissors");
+                string userChoice = Console.ReadLine();
```
breaman: could have simplified this down to just "return Console.ReadLine();"

Assignment3/PaperRockScissors/PaperRockScissors/PaperRockScissorsClass.cs
```diff
@@ -0,0 +1,168 @@
...
+        public static Tuple<string, int> GameAction()
+        {
+            int PlayerScore = 100, ComputerScore = 100;
+
+            while (PlayerScore > 0 && ComputerScore > 0)
+            {
+                String userChoice = GetUserInput();
+                int computerChoice = GetComputerInput();
+
+                if (computerChoice == 1)
```
breaman: This logic should have been pulled out into a different method so that you could have tested the logic. Unfortunately by doing things this way, the bulk of your program isn't tested, which isn't good.

breaman: It would be good to know what 1, 2, and 3 were regarding the computer's move. Things are confusing here because even though computerChoice was 1 in this if block, you have "The computer chose rock", "The computer chose paper", and "The computer chose scissors". My guess is 1 can't be all 3. It appears you really have it set up so that if the computer chose 1, then it would always be a tie, a 2 would be a win for the computer, and 3 is win for the player, but that really isn't the "computerChoice".

PR #57, opened 2018-10-19
Assignment4 Submission

KennethWhite:


Assignment4/Assignment4/UniversityCourse.cs
```diff
@@ -0,0 +1,115 @@
...
+                if (value.Length < 3 || string.IsNullOrWhiteSpace(value))
+                {
+                    throw new ArgumentException($"Department code must be at least 3 characters {value}");
+                }
+                _Department = value;
+
+            }
+        }
+        private string _Department;
+
```
NikolasGraham: Do you want to be able to set course ID, or rather have it as an immutable number so it can't be accidentally changed later?

PR #58, opened 2018-10-19
Jacob Blazina - Assignment4 - PR 1

blazinaj:


Assignment4/Assignment4/CourseDatabase.cs
```diff
@@ -0,0 +1,47 @@
...
+        public void AddCourse(Course course)
+        {
+            CourseList.Add(course);
+        }
+        public int CourseCount
+        {
+            get { return CourseList.Count; }
+        }
+
+        public void Display(string param = "all")
```
MarkMichaelis: No, this does not leverage polymorphism or pattern matching.
I would have expected something like this:
`
        public string Display(object @object)
        {
            switch (@object)
            {
                case Course course:
                    return course.GetSummaryInformation();
                default:
                    return @object.ToString();
                case null:
                    throw new ArgumentNullException();                   
            }
        }
`

Assignment4/Assignment4/CourseDatabase.cs
```diff
@@ -0,0 +1,47 @@
...
+    {
+        public List<Course> CourseList = new List<Course>();
+
+        public static void Main(string[] args)
+        {
+            Console.WriteLine("University Database");
+
+        }
+
+        public void AddCourse(Course course)
```
MarkMichaelis: Your `CourseList` field is public, exposing CourseList.Add() so there is little point in this method.

Assignment4/Assignment4/CourseDatabase.cs
```diff
@@ -0,0 +1,47 @@
+﻿using System;
+using System.Collections.Generic;
+
+namespace Assignment4
+{
+    public class CourseDatabase
+    {
+        public List<Course> CourseList = new List<Course>();
```
MarkMichaelis: Fields should never be public.  This is especially true for collections (which should also generally be read only).

Assignment4/Assignment4/CourseDatabase.cs
```diff
@@ -0,0 +1,47 @@
+﻿using System;
+using System.Collections.Generic;
+
+namespace Assignment4
+{
+    public class CourseDatabase
```
MarkMichaelis: I think your approach of having a CourseDatabase is good idea.

Assignment4/Assignment4/Course.cs
```diff
@@ -0,0 +1,27 @@
...
+
+        public Course(string name)
+        {
+            CourseName = name;
+        }
+
+     
+        public virtual string GetSummaryInformation()
+        {
+            string info = "No information available";
```
MarkMichaelis: Why?  You at least have a name and a course type.

Assignment4/Assignment4/SingleCourse.cs
```diff
@@ -0,0 +1,32 @@
...
+using System.Text;
+
+namespace Assignment4
+{
+    public class SingleCourse : Course
+    {
+        public DateTime Date { get; set; }
+
+        public SingleCourse(string name)
+            : base(name)
```
MarkMichaelis: Nice job invoking the base class constructor.

Assignment4/Assignment4/Course.cs
```diff
@@ -0,0 +1,27 @@
...
+using System.Collections.Generic;
+using System.Text;
+
+namespace Assignment4
+{
+    public class Course
+    {
+        public Type CourseType { get; set; }
+
+        public string CourseName { get; set; }
```
MarkMichaelis: You don't have any validation.

Assignment4/Assignment4/Course.cs
```diff
@@ -0,0 +1,27 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Assignment4
+{
+    public class Course
+    {
+        public Type CourseType { get; set; }
```
MarkMichaelis: Course type is surely not a type-of-dotnet object.  I expect an enum might be appropriate but not System.Type.

Assignment4/Assignment4.Tests/CourseDatabase.Tests.cs
```diff
@@ -0,0 +1,48 @@
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using System;
+
+namespace Assignment4.Tests
```
MarkMichaelis: You have a significant shortage of tests with only 44% of your code covered, very little of which is boundary testing.

Assignment4/Assignment4/Course.cs
```diff
@@ -0,0 +1,27 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Assignment4
+{
+    public class Course
```
MarkMichaelis: There doesn't seem to be a count on how many instances are created - on any of your classes.

PR #59, opened 2018-10-19
Assignment4 submission.

joeriddles:


Assignment4/src/CalendarItems/Course.cs
```diff
@@ -0,0 +1,93 @@
...
+
+			if (!DateTime.TryParse(startTime, out DateTime result))
+				throw new ArgumentException("Start Time invalid.");
+
+			if (!daysOfWeek.Any())
+				throw new ArgumentException("Days of week invalid.");
+
+			if (DateTime.Compare(startDate, endDate) > 0)
+				throw new ArgumentException("Start date is after end Date");
+
```
afleiger: It might be better to handle all of the parameter checks inside the property logic rather than inside the constructor, Especially for StartDate and EndDate.

Assignment4/src/CalendarItems/Application.cs
```diff
@@ -0,0 +1,23 @@
...
+				case Course course:
+					return course.GetSummaryInformation();
+				case Event evnt:
+					return evnt.GetSummaryInformation();
+				case CalendarItem ci:
+					return ci.GetSummaryInformation();
+				case Object objct:
+					return objct.ToString(); 
+				default: throw new ArgumentException();
+			}
```
afleiger: Since Course and Event both derive from CalendarItem, why not remove the first 2 case statements and allow the CalendarItem case to catch both and use polymorphic behavior to call GetSummaryInformation()?

Assignment4/src/CalendarItems/Course.cs
```diff
@@ -0,0 +1,93 @@
...
+		public DateTime EndDate { get; set; }
+
+		private TimeSpan timeSpan;
+		public TimeSpan TimeSpan {
+			get { return timeSpan; }
+			set { timeSpan = EndDate - StartDate; }
+		} 
+
+		private readonly Regex courseIdRegex = new Regex("^[a-zA-Z]{4}[0-9]{3}$");
+		public Regex CourseIdRegex => courseIdRegex;
```
afleiger: Nice use of regular expressions for ensuring valid CourseID.

PR #61, opened 2018-10-20
Assignment4

toha87:


UniversityCourseWork/UniversityCourseWork/UniversityCourse.cs
```diff
@@ -0,0 +1,83 @@
...
+
+namespace UniversityCourseWork
+{
+    internal class UniversityCourse
+    {
+        public static int COUNT;
+
+        private readonly int _CourceId;
+        public int CourseId { get; }
+
```
Nematodes: It seems that _CourceId is set by the constructor but never used, while CourseId is not set by the constructor but is used by Deconstruct. This results in CourseId being set to its default value of 0.

Nematodes: If the auto-implemented property CourseId is created with only a get method, then its backing field is treated as a public read-only value. Does the _CourceId field have a reason to exist?

UniversityCourseWork/UniversityCourseWork/UniversityCourse.cs
```diff
@@ -0,0 +1,83 @@
...
+        public void AddCourse(UniversityCourse course)
+        {
+            CourseList.Add(course);
+        }
+        public int CourseCount(List<UniversityCourse> CourseList)
+        {
+            return CourseList.Count;
+        }
+
```
Nematodes: Was your intention really to count the number of elements in the passed-in list?

Doing so seems unnecessary. Perhaps you were trying to count the number of elements in the current UniversityCourse's list?

UniversityCourseWork/UniversityCourseWorkUnitTest/UniversityUnitTest.cs
```diff
@@ -0,0 +1,96 @@
...
+        public void Create_ListOf_UniversityCourses()
+        {
+            UniversityCourse FirstCourse = new UniversityCourse(129, 302, "Computing Ethics");
+            UniversityCourse.COUNT++;
+            UniversityCourse SecondCourse = new UniversityCourse(145, 371, ".NET Programming");
+            UniversityCourse.COUNT++;
+
+            Course.AddCourse(FirstCourse);
+            Course.AddCourse(SecondCourse);
+
```
Nematodes: It initially confused me when you referred to the property Course before declaring it later in the file.

UniversityCourseWork/UniversityCourseWork/UniversityCourse.cs
```diff
@@ -0,0 +1,83 @@
...
+{
+    internal class UniversityCourse
+    {
+        public static int COUNT;
+
+        private readonly int _CourceId;
+        public int CourseId { get; }
+        
+        public int CourseNumber { get; set; }
+
```
Nematodes: Are course numbers allowed to be negative values?

UniversityCourseWork/UniversityCourseWorkUnitTest/UniversityUnitTest.cs
```diff
@@ -0,0 +1,96 @@
...
+            UniversityCourse.COUNT++;
+        }
+
+        [TestMethod]
+        public void TestScheduleClass()
+        {
+            Schedule schedule = new Schedule(125, 320, "Algorithms", "10.17.2018");
+            UniversityCourse.COUNT++;
+
+        }
```
Nematodes: This test appears to be incomplete, and the name of the test method is also nondescriptive.

UniversityCourseWork/UniversityCourseWork/UniversityCourse.cs
```diff
@@ -0,0 +1,83 @@
...
+        public static int Display(object @object)
+        {
+            switch (@object)
+            {
+                case Schedule course:
+                    CourseList.Add(course);
+                    break;
+                case UniversityCourse course:
+                    CourseList.Add(course);
+                    break;
```
Nematodes: Why does a method called "Display" add to the course list?

PR #62, opened 2018-10-20
Assignment4

NikolasGraham:


Assignment4/src/UniversityCourseWork/UniversityCourse.cs
```diff
@@ -0,0 +1,18 @@
+﻿using System;
+
+namespace UniversityCourseWork
+{
+    public class UniversityCourse
+    {
+
+        public string Course_Schedule = "Algebra.Monday.8AM"
```
JeffHowes: Since the Course_Schedule string is not used outside this class, the scope could be reduced to private.

JeffHowes: The C# book's coding guidelines states that PascalCase and camelCase are the two styles for casing.  The snake case in "Course_Schedule" is very easy to read, but it is outside the 2 recommended C# casing styles.

Assignment4/src/UniversityCourseWork/MainClass.cs
```diff
@@ -0,0 +1,104 @@
...
+            get
+            {
+                return degree;
+            }
+            set
+            {
+                if (value < _minDegree || value > _maxDegree)
+                {
+                    throw new ArgumentOutOfRangeException();
+                }
```
JeffHowes: I was expecting to see a else statement assigning value to degree in the Degree properties set section.
```set
            {
                if (value < _minDegree || value > _maxDegree)
                {
                    throw new ArgumentOutOfRangeException();
                }
}```

File: Assignment4/src/UniversityCourseWork/MainClass.cs [hunk 3]
```diff
@@ -0,0 +1,104 @@
...
+        {
+            get
+            {
+                return _StringToChange;
+            }
+        }
+
+        private readonly int _minDegree = -42;
+        private readonly int _maxDegree = 42;
+        public int degree;
```
JeffHowes: The backing field "degree" should be "_Degree", for consistency.

Assignment4/src/UniversityCourseWork/MainClass.cs
```diff
@@ -0,0 +1,104 @@
...
+        private List<string> eventDays = new List<string>();
+        private List<string> courseTimes = new List<string>();
+        private List<string> eventTimes = new List<string>();
+
+        public void CourseCollection()
+        {
+            UniversityCourse UC = new UniversityCourse();
+            Event EV = new Event();
+
+            string courses = UC.GetSummaryInformation();
```
JeffHowes: Can you explain why you decided to split the course's and event's List, Days, and Times into different lists?

Assignment4/src/UniversityCourseWork/MainClass.cs
```diff
@@ -0,0 +1,104 @@
...
+            string[] eventParts = events.Split(".");
+
+            for(int i = 0; i < courseParts.Length; i+=3)
+            {
+                courseList.Add(courseParts[0]);
+                courseDays.Add(courseParts[1]);
+                courseTimes.Add(courseParts[2]);
+
+                eventList.Add(eventParts[0]);
+                eventDays.Add(eventParts[1]);
```
JeffHowes: The for loop has a dependency that there must be an equal number of courses to events,  Which is OK if your code can ensure there is never more of one than the other.

Assignment4/src/UniversityCourseWork/MainClass.cs
```diff
@@ -0,0 +1,104 @@
...
+        private int _CalculatedProperty;
+        public object CalculatedProperty
+        {
+            get
+            {
+                return _CalculatedProperty * _CalculatedProperty;
+            }
+            set
+            {
+                _CalculatedProperty = (int)value;
```
MarkMichaelis: What if value is not an int?  Having a getter that returns a different value in the getter would be very surprising to the developer invoking the property.

Assignment4/src/UniversityCourseWork/Event.cs
```diff
@@ -0,0 +1,19 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace UniversityCourseWork
+{
+    public class Event
```
MarkMichaelis: No.. hard coding the values defeats the purpose of a class - essentially making limited for a single instance and nothing else.

Assignment4/src/UniversityCourseWork/UniversityCourseWork.csproj
```diff
@@ -0,0 +1,8 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
```
MarkMichaelis: See [Assignment4/GradingNotes.md #22](https://github.com/IntelliTect-Samples/EWU-CSCD371-2018-Fall/blob/master/Assignment4/GradingNotes.md#22-should-all-projects-be-of-outputtype-exe)

PR #32, opened 2018-10-05
Assignment2

toha87:

PR #64, opened 2018-10-20
Assignment4

valdynhunt:


Assignment4/UniversityCourse/src/UniversityCourse/UniversityCourse.cs
```diff
@@ -0,0 +1,279 @@
+﻿using System;
+using System.Text;
+
+namespace UniversityCourse
+{
+    public class UniversityCourse : Event
+    {
+       
+
+        private int _Crn;
```
MarkMichaelis: See [Assignment4/GradingNotes.md #32](https://github.com/IntelliTect-Samples/EWU-CSCD371-2018-Fall/blob/master/Assignment4/GradingNotes.md#32-is-it-okay-to-return-from-a-property-setter-without-changing-the-value-even-though-a-value-was-specified-that-was-different-than-the-current-value)

PR #18, opened 2018-09-27
Andrew Fleiger Assignment 1 PR

afleiger:
This is the files for Andrew Fleiger's Assignment 1.

PR #66, opened 2018-10-21
Assignment4

JeffHowes:
Committing Assignment4


Assignment4/HW4/Course.cs
```diff
@@ -0,0 +1,112 @@
...
+        {
+            return "COURSE: " + CourseNumber + "| " + Name + "| " + InstructorLastName
+                + ", " + InstructorFirstName + "| " + CourseSchedule;
+        }
+
+        //This deconstructor was adapted from the Lecture code posted to Git.
+        private (string FirstName, string LastName) ParseName(string value)
+        {
+            int separatorIndex = value.IndexOf('.');
+            if (separatorIndex < 2 || separatorIndex > value.Length - 2)
```
bcaesar42: `separatorIndex < 2` should be `separatorIndex < 1`. This part of the boolean statement is meant to ensure that there is at least one character preceding the period. However, `seperatorIndex < 2` will require that there be at least two characters preceding the period. For example, passing "J.Howes"into the method will result in an exception, because the period is at index 1.

Assignment4/HW4/Course.cs
```diff
@@ -0,0 +1,112 @@
...
+            {
+                throw new ArgumentException(nameof(InstructorUserName),
+                    "InstructorUserName must be of the format <FirstName>.<LastName>");
+            }
+            else
+            {
+                return (FirstName: value.Substring(0, separatorIndex),
+                    LastName: value.Substring(separatorIndex + 1,
```
bcaesar42: `value.Substring(separatorIndex + 1, value.Length - separatorIndex - 1)` can be simplified as `value.Substring(separatorIndex + 1)`. The single-parameter version of Substring will start at the specified index and go to the end of the string, instead of manually specifying the end index.

Assignment4/HW4/UniversityCalander.cs
```diff
@@ -0,0 +1,31 @@
...
+    public abstract class UniversityCalender
+    {
+        public static int TotalCalendarItems = 0; //inc this total from within the course and event constructors
+        public string Location { get; set; }
+        public string StartDate { get; set; }
+        public string Name { get; set; }
+
+        public abstract string Display();     
+        
+        public void incTotalCalendarItems()
```
bcaesar42: Instead of relying on each of the child classes to call this method in each of their constructors, you could just call this method inside of the constructor for this base class.

Assignment4/HW4UnitTest/CourseTests.cs
```diff
@@ -0,0 +1,88 @@
...
+        }
+
+        [TestMethod]
+        [DataRow(11)]
+        [DataRow(37)]
+        [ExpectedException(typeof(ArgumentException))]
+        public void StudentCount_11Students_Fail(int numOfStudents)
+        {
+            course.StudentCount = numOfStudents;
+            //Assert.AreEqual(33, course.StudentCount);
```
bcaesar42: Unused line of code, consider removing it.

Assignment4/HW4/UniversityCalander.cs
```diff
@@ -0,0 +1,31 @@
...
+        {
+            TotalCalendarItems++;
+        }
+
+        public int getTotalCalendarItems()
```
MarkMichaelis: See [Assignment4/GradingNotes.md #27](https://github.com/IntelliTect-Samples/EWU-CSCD371-2018-Fall/blob/master/Assignment4/GradingNotes.md#27-given-a-getpropertyname-and-a-setpropertyname-methods-should-refactoring-into-a-property-be-used)

Assignment4/HW4/UniversityCalander.cs
```diff
@@ -0,0 +1,31 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace HW4
+{
+    public abstract class UniversityCalender
+    {
+        public static int TotalCalendarItems = 0; //inc this total from within the course and event constructors
+        public string Location { get; set; }
```
MarkMichaelis: See [Assignment4/GradingNotes.md #18](https://github.com/IntelliTect-Samples/EWU-CSCD371-2018-Fall/blob/master/Assignment4/GradingNotes.md#18-is-it-okay-to-have-properties-that-are-reference-types-and-allow-null)

Assignment4/HW4/Event.cs
```diff
@@ -0,0 +1,30 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace HW4
+{
+    public class Event : UniversityCalender
+    {
+
+        public Event(String name, String location, String startDate)
```
MarkMichaelis: See [Assignment4/GradingNotes.md #28](https://github.com/IntelliTect-Samples/EWU-CSCD371-2018-Fall/blob/Assignment-4-Grading/Assignment4/GradingNotes.md#28-which-is-preferable-string-the-key-word-or-string-the-net-type)

Assignment4/HW4/Event.cs
```diff
@@ -0,0 +1,30 @@
...
+        {
+            Name = name;
+            Location = location;
+            StartDate = startDate;
+            incTotalCalendarItems();
+        }
+        
+        public override string Display()
```
MarkMichaelis: It would seem to have been significantly simpler to just put `GetSummaryInformation()` in the `UniversityCalendar` class and leverage that for polymorphism.  Why add another method that calls `GetSummaryInformation()`.  It isn't as though there was a particular interface that needed to be supported.

Assignment4/HW4/UniversityCalander.cs
```diff
@@ -0,0 +1,31 @@
...
+        public string StartDate { get; set; }
+        public string Name { get; set; }
+
+        public abstract string Display();     
+        
+        public void incTotalCalendarItems()
+        {
```
MarkMichaelis: Why have this method at all?  Couldn't you just call TotalCalendarItems from the derived classes?

Assignment4/HW4/UniversityCalander.cs
```diff
@@ -0,0 +1,31 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace HW4
+{
+    public abstract class UniversityCalender
+    {
+        public static int TotalCalendarItems = 0; //inc this total from within the course and event constructors
```
MarkMichaelis: See [Assignment4/GradingNotes.md #29](https://github.com/IntelliTect-Samples/EWU-CSCD371-2018-Fall/blob/master/Assignment4/GradingNotes.md#29-is-it-okay-to-have-public-fields)

Assignment4/HW4/Course.cs
```diff
@@ -0,0 +1,112 @@
...
+
+        public int InstructorSalory
+        {
+            get
+            {
+                return _InstructorSalory;
+            }
+            set
+            {
+                _InstructorSalory = value + (StudentCount / 10) * 1000;
```
MarkMichaelis: See [Assignment4/GradingNotes.md #7](https://github.com/IntelliTect-Samples/EWU-CSCD371-2018-Fall/blob/master/Assignment4/GradingNotes.md#7-is-it-okay-to-change-the-value-assigned-in-the-setter-of-a-property)

Assignment4/HW4/Course.cs
```diff
@@ -0,0 +1,112 @@
...
+        }
+
+        public string GetSummaryInformation()
+        {
+            return "COURSE: " + CourseNumber + "| " + Name + "| " + InstructorLastName
+                + ", " + InstructorFirstName + "| " + CourseSchedule;
+        }
+
+        //This deconstructor was adapted from the Lecture code posted to Git.
+        private (string FirstName, string LastName) ParseName(string value)
```
MarkMichaelis: See [Assignment4/GradingNotes.md #26](https://github.com/IntelliTect-Samples/EWU-CSCD371-2018-Fall/blob/master/Assignment4/GradingNotes.md#26-what-is-the-correct-signature-of-a-deconstructor)

Assignment4/HW4/Course.cs
```diff
@@ -0,0 +1,112 @@
...
+        {
+            int separatorIndex = value.IndexOf('.');
```
MarkMichaelis: I recommend using `string.Split()` instead.  It would make this method simpler I expect.

Assignment4/HW4/Course.cs
```diff
@@ -0,0 +1,112 @@
...
+            return GetSummaryInformation();
+        }
+
+        public string GetSummaryInformation()
+        {
+            return "COURSE: " + CourseNumber + "| " + Name + "| " + InstructorLastName
+                + ", " + InstructorFirstName + "| " + CourseSchedule;
+        }
+
+        //This deconstructor was adapted from the Lecture code posted to Git.
```
MarkMichaelis: See [Assignment4/GradingNotes.md #28](https://github.com/IntelliTect-Samples/EWU-CSCD371-2018-Fall/blob/Assignment-4-Grading/Assignment4/GradingNotes.md#30-should-methods-that-dont-access-any-instance-data-be-static)

Assignment4/HW4/Course.cs
```diff
@@ -0,0 +1,112 @@
...
+            {
+                return _CourseNumber;
+            }
+        }
+        private string _CourseNumber;
+        
+        public string CourseSchedule { get; set; }
+        public string InstructorFirstName { get; set; }
+        public string InstructorLastName { get; set; }
+        public string InstructorUserName
```
MarkMichaelis: Nice job with this calculated property.

Assignment4/HW4/Course.cs
```diff
@@ -0,0 +1,112 @@
...
+{   
+    public class Course : UniversityCalender
+    {
+        public Course(string name, string courseNumber, string courseSchedule, string instructorUserName)
+        {
+            _CourseNumber = courseNumber;
+            Name = name;
+            CourseSchedule = courseSchedule;
+            InstructorUserName = instructorUserName;
+            incTotalCalendarItems();
```
MarkMichaelis: This tracks the total number of `UniversityCalendar` instances.  What tracks `Course` or `Event` instances?

Assignment4/HW4/Course.cs
```diff
@@ -0,0 +1,112 @@
...
+                + ", " + InstructorFirstName + "| " + CourseSchedule;
+        }
+
+        //This deconstructor was adapted from the Lecture code posted to Git.
+        private (string FirstName, string LastName) ParseName(string value)
+        {
+            int separatorIndex = value.IndexOf('.');
+            if (separatorIndex < 2 || separatorIndex > value.Length - 2)
+            {
+                throw new ArgumentException(nameof(InstructorUserName),
```
MarkMichaelis: Using the argument name of `InstructorUserName` when there is no such argument is misleading.  I realize that this is because the property is invoked from InstructorUserName but if the method happened to be called from elsewhere, it wouldn't be correct.  It isn't valid to say the argument name is something that doesn't exist as an argument to the method throwing the exception.

Assignment1/src/HelloWorld/Program.cs
```diff
@@ -0,0 +1,14 @@
+﻿using System;
+
+namespace HelloWorld
+{
+    public class Program
```
MarkMichaelis: Given only static members this class should be a static class.

PR #25, opened 2018-10-04
Assignment 2 Submission

Nematodes:
This is the baseline calculator application.

It handles basic arithmetic operators, negative numbers, and most edge cases (all of the ones I could find with a moderate amount of effort.)

It also contains comprehensive arithmetic unit tests, as well as the string unit tests.

I may (unlikely) update the calculator's functionality in a later commit.

PR #68, opened 2018-10-21
Assignment4

Chester-Southwood:
Homework Submission for Assignment4.


Assignment4/Assignment4/tests/CourseTest.cs
```diff
@@ -0,0 +1,124 @@
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using System;
+
+namespace tests
```
KennethWhite: You could be a little more descriptive with the namespace here.

Assignment4/Assignment4/src/Course.cs
```diff
@@ -0,0 +1,116 @@
...
+            }
+            set
+            {
+                if(value.Length > 0)
+                {
+                    _Subject = value;
+                }
+                else if(value.Length == 0)
+                {
+                    throw new NotSupportedException("Subject cannot be empty.");
```
KennethWhite: I don't think NotSupportiveException is a good choice here. One of the ArgumentExceptions might be more descriptive, because this class _does_ support assigning into the subject
```suggestion
                    throw new ArgumentException("Subject cannot be empty.");           
```

Assignment4/Assignment4/tests/CourseTest.cs
```diff
@@ -0,0 +1,124 @@
...
+        {
+            Course.RoomNum = 42;
+            Assert.Fail();
+        }
+
+        [TestMethod]
+        public void TestValid_SetRoomNum()
+        {
+            Course.RoomNum = 123;
+            int actual = Course.RoomNum;
```
KennethWhite: You dont really need to store the RoomNum into a temporary variable.
```suggestion
             Assert.AreEqual(123, Course.RoomNum)
```

Assignment4/Assignment4/tests/CourseTest.cs
```diff
@@ -0,0 +1,124 @@
...
+    [TestClass]
+    public class TestCourse
+    {        
+        private static Assignment4.Course Course { get; set; }
+
+        [TestInitialize]
+        public void SetupUnivCourse()
+        {
+            Course = new Assignment4.Course("Computer Science", ".net", 208, 21, 2, 4);
+            Course.resetCourseTotal();
```
KennethWhite: Do you really want to reset the course total after creating the course? It seems you should reset the total before making the new test.

Assignment4/Assignment4/tests/CourseTest.cs
```diff
@@ -0,0 +1,124 @@
...
+        {
+            int actual = Course.FloorNum;
+            Assert.AreEqual(2, actual);
+        }
+
+        [TestMethod]
+        public void TestValid_GetCourseTotalInitial()
+        {
+            int actual = Course.CourseTotal;
+            Assert.AreEqual(0, actual);
```
KennethWhite: As noted above in your [TestInitialize] method, while technically this test passes it probably shouldn't. Before you call this test method you create a new course, why then would you expect 0 on this line if we just made one? This could be corrected by moving the resetCourseTotal() call in the initialize method to before the course instantiation.

Assignment4/Assignment4/tests/CourseTest.cs
```diff
@@ -0,0 +1,124 @@
...
+        }
+
+
+        [TestMethod]
+        public void TestValid_GetCourseTotalMultipleCourse()
+        {
+            Assignment4.Course t = new Assignment4.Course("cs", "c and unix", 204, 2, 3, 4);
+            Assignment4.Course t2 = new Assignment4.Course("cs", "c and unix", 204, 2, 3, 4);
```
KennethWhite: As noted in the prior test method, if we have a static variable on line 9 that is instantiated before every test, _and_ we create two more, I would expect 3 to be the tested value here.

Assignment4/Assignment4/tests/EventTest.cs
```diff
@@ -0,0 +1,99 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace tests
+{
+    [TestClass]
+    public class Test
```
KennethWhite: This class name should probably better indicate what will be tested.

Assignment4/Assignment4/tests/EventTest.cs
```diff
@@ -0,0 +1,99 @@
...
+    {
+        private static Assignment4.Event Event { get; set; }
+
+        [TestInitialize]
+        public void SetupUnivCourse()
```
KennethWhite: This method name indicates we are setting up a UniversityCourse rather than an Event. You don't have any other classes or methods referencing UniversityCourse, only Course.

Assignment4/Assignment4/src/Course.cs
```diff
@@ -0,0 +1,116 @@
+using System;
```
MarkMichaelis: See [Assignment4/GradingNotes.md #24](https://github.com/IntelliTect-Samples/EWU-CSCD371-2018-Fall/blob/master/Assignment4/GradingNotes.md#24-is-it-okay-to-have-extra-using-directives-even-if-they-arent-ever-used)

Assignment4/Assignment4/src/Display.cs
```diff
@@ -0,0 +1,36 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Assignment4
+{
+    public static class DisplayClass
+    {
+        public static string Display(Object e)
```
MarkMichaelis: See [Assignment4/GradingNotes.md #25](https://github.com/IntelliTect-Samples/EWU-CSCD371-2018-Fall/blob/Assignment-4-Grading/Assignment4/GradingNotes.md#25-is-it-okay-to-have-parameter-names-of-a-single-letter)

Assignment4/Assignment4/src/Display.cs
```diff
@@ -0,0 +1,36 @@
...
+                case Event eventTemp:
+                    return eventTemp.GetSummaryInformation();
+                case null:
+                    throw new ArgumentNullException("Cannot display a null object");
+                default:
+                    return e.ToString();
+            }
+        }
+
+        public static string DisplayList(List<Event> list)
```
MarkMichaelis: Why a separate method?  Why not just have this as part of your `Display()` method?  If you need to refactor (likely) then call a method that includes this logic.  Preferably, however, don't relay on the programmer to know there are two separate methods.

Assignment4/Assignment4/tests/DisplayTest.cs
```diff
@@ -0,0 +1,60 @@
...
+            Assert.AreEqual(Display(Event), Event.GetSummaryInformation());
+        }
+
+        [TestMethod]
+        public void TestValid_Display()
+        {
+            Assert.AreEqual(Display(Course), Course.GetSummaryInformation());
+        }
+
+        [TestMethod]
```
MarkMichaelis: Nice use of DataRow

Assignment4/Assignment4/tests/DisplayTest.cs
```diff
@@ -0,0 +1,60 @@
...
+        public void TestValid_SummaryAndToString(object e)
+        {
+            Assert.AreEqual(e.ToString(), Display(e));
+        }
+
+
+        [TestMethod]
+        public void TestValid_SummaryList()
+        {
+            StringBuilder expected = new StringBuilder();
```
MarkMichaelis: This is a good use of `StringBuilder`

Assignment4/Assignment4/tests/EventTest.cs
```diff
@@ -0,0 +1,99 @@
...
+            Assert.AreEqual(count + 1, Event.EventNum);
+        }
+
+        [TestMethod]
+        public void TestValid_Deconstruct()
+        {
+            int tempStartHour = 4;
+            int tempEndHour = 3;
+            String tempTitle = "School";
+            Event.Deconstruct(out tempStartHour, out tempEndHour, out tempTitle);
```
MarkMichaelis: See [Assignment4/GradingNotes.md #15](https://github.com/IntelliTect-Samples/EWU-CSCD371-2018-Fall/blob/Assignment-4-Grading/Assignment4/GradingNotes.md#15-how-do-you-invoke-a-deconstrutor)

Assignment4/Assignment4/tests/EventTest.cs
```diff
@@ -0,0 +1,99 @@
...
+            Event = new Assignment4.Event(5, 7, "english");
+            Assert.AreEqual(count + 1, Event.EventNum);
+        }
+
+        [TestMethod]
+        public void TestValid_Deconstruct()
+        {
+            int tempStartHour = 4;
+            int tempEndHour = 3;
+            String tempTitle = "School";
```
MarkMichaelis: While not technically wrong, I recommend favoring lowercase `string`.  Regardless, you should use the same convention everywhere and you have both `string` and `String`.

Assignment4/Assignment4/tests/EventTest.cs
```diff
@@ -0,0 +1,99 @@
...
+            Assert.AreEqual(22, Event.StartHour);
+            Assert.AreEqual(23, Event.EndHour);
+        }
+
+        [TestMethod]
+        [ExpectedException(typeof(NotSupportedException))]
+        public void TestInvalid_EventTime()
+        {
+            Event = new Assignment4.Event(22, 24, "woodshop");
+            Assert.Fail();
```
MarkMichaelis: Calling Fail() should never get called so why have it?

Assignment4/Assignment4/src/Course.cs
```diff
@@ -0,0 +1,116 @@
...
+        }
+
+        public int CourseTotal 
+        {
+            get
+            {
+                int temp = _CourseTotal;
+                return temp;
+            }
+            private set
```
MarkMichaelis: Good choice having a private accessibility modifier here.

PR #72, opened 2018-10-31
Assignment5

JeffHowes:
Creating pull request for Assignment 5


Assignment1/HelloWorld.sln
```diff
@@ -0,0 +1,59 @@
...
+		HideSolutionNode = FALSE
+	EndGlobalSection
+	GlobalSection(NestedProjects) = preSolution
+		{469FD78A-7FAA-4610-B6A0-D503E9C4F0F4} = {705324DB-2D59-46D6-B73E-785945143F07}
+		{44054BC3-E789-45D4-BBDB-9E800FA02695} = {8E6204A7-53A8-427E-801E-E830866D2E92}
+	EndGlobalSection
+	GlobalSection(ExtensibilityGlobals) = postSolution
+		SolutionGuid = {8D67E670-5C03-480D-8A24-47DBB8FB61E2}
+	EndGlobalSection
+EndGlobal
```
Keboo: It looks like you still have your assignment 1 files included in your history chain. If you have time I am in class 11/8 and 11/13. Let's discuss and address it then.

Assignment5/HW5.Tests/CourseTests.cs
```diff
@@ -0,0 +1,88 @@
...
+        public void CleanupTest()
+        {
+            course = null;
+        }
+
+
+        [TestMethod]
+        public void InstructorUserName_AssignTomDotCapal_Success()
+        {
+            course.InstructorUserName = "Tom.Capal";
```
Keboo: Not a big deal but it is spelled "Capaul"

Assignment5/HW5/Program.cs
```diff
@@ -0,0 +1,146 @@
...
+                usersMenuChoice = DisplayUserMenu();
+
+                if(usersMenuChoice == (int)MenuSelection.CreateEvent)
+                {
+                    EventList.Add(UserCreatesEvent() );
+                }
+
+                if (usersMenuChoice == (int)MenuSelection.DisplayEvents)
+                {
+                    System.Console.WriteLine("Printing out all events...");
```
Keboo: Is there a reason these calls are using `System.Console.WriteLine` rather than going through your `IConsole` abstraction?

Assignment5/HW5/Program.cs
```diff
@@ -0,0 +1,146 @@
...
+            while(true)
+            {
+                Console.WriteLine("\nUser Menu:");
+                Console.WriteLine("Press 1 to create a new Calander Event");
+                Console.WriteLine("Press 2 to display all Calander Events");
+                Console.WriteLine("Press 3 to check all events for room conflicts - (uses IEvent interface)");
+                Console.WriteLine("Press 4 to public annouce all events - (uses ext. method)");
+                Console.WriteLine("Press 5 to exit");
+
+                string input = System.Console.ReadLine();
```
Keboo: Shouldn't this be using `Console.ReadLine()` rather than going directly to `System.Console.ReadLine()`?

Assignment5/HW5/Program.cs
```diff
@@ -0,0 +1,146 @@
...
+
+        public static void AnnounceAllEventsNames(List<UniversityCalender> eventList)
+        {
+            foreach (UniversityCalender UnivCalEvent in eventList)
+            {
+                ((Event)UnivCalEvent).Name.PublicAnnoucement();
+            }
+        }
+        
+        public static void DisplayAllEvents(Object obj)
```
Keboo: Favor using the keyword `object` over the type name `Object`

Assignment5/HW5/Program.cs
```diff
@@ -0,0 +1,146 @@
...
+            switch (obj)
+            {
+                case Course course:
+                    System.Console.WriteLine(course.GetSummaryInformation());
+                    break;
+                case Event @event:
+                    System.Console.WriteLine(@event.GetSummaryInformation());
+                    break;
+                case null:
+                    throw new ArgumentNullException(nameof(obj));
```
Keboo: Good use of pattern matching switch statement!

Assignment5/HW5/RealConsole.cs
```diff
@@ -0,0 +1,28 @@
...
+
+            System.Console.Write(line);
+        }
+
+        public string ReadLine()
+        {
+            return Console.ReadLine();
+        }
+    }
+}
```
Keboo: Good work on implementing the interface. There is a very minor inconsistency here. In `ReadLine()` you simply use the class name (`Console`), while you use the fully qualified named (`System.Console`) in the other methods.

It also does not appear that you even instantiate this class.

Assignment5/HW5/TestableConsole.cs
```diff
@@ -0,0 +1,26 @@
...
+using System.Text;
+using HW5;
+
+namespace HW5
+{
+    public class TestableConsole : IConsole
+    {
+        public string LastWrittenLine { get; private set; }
+        public string LastWrite { get; private set; }
+        public string LastReadLine { private get; set; }
```
Keboo: Typically you don't have properties with a setter that is more accessible than the getter. I realize in this case the getter is only expected to be called by the `ReadLine()` method, but it makes the API cleaner if both the getter and setter are public.

Assignment5/HW5/TestableConsole.cs
```diff
@@ -0,0 +1,26 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+using HW5;
+
+namespace HW5
+{
+    public class TestableConsole : IConsole
```
Keboo: Nice work on implementing a test implementation for the interface

Assignment5/HW5/UniversityCalander.cs
```diff
@@ -0,0 +1,31 @@
...
+        {
+            TotalCalendarItems++;
+        }
+
+        public int getTotalCalendarItems()
+        {
+            return TotalCalendarItems;
+        }
+
+        public void setTotalCalendarItems(int num)
```
Keboo: Naming issue for most of the methods names in this file. Should be CamelCase.

Assignment5/HW5.Tests/CourseTests.cs
```diff
@@ -0,0 +1,88 @@
...
+        [DataRow(37)]
+        [ExpectedException(typeof(ArgumentException))]
+        public void StudentCount_11Students_Fail(int numOfStudents)
+        {
+            course.StudentCount = numOfStudents;
+            //Assert.AreEqual(33, course.StudentCount);
+        }
+
+        [TestMethod]
+        public void InstructorSalory_SalaryOf20000And33Students_Success()
```
Keboo: Typo "Salory"

Assignment5/HW5/Program.cs
```diff
@@ -0,0 +1,146 @@
+﻿using System;
+using System.Collections.Generic;
+using HW5;
+
+namespace HW5
+{
+    public class Program 
+    {
+        public static IConsole Console { get; set; }
```
Keboo: This type of "injection" is called Property Injection. I see that you properly set this for unit testing, however I don't see it getting set for the normal running of the application. This causes your app to throw a NullReferenceException when run. I would have expected to see something like:
```C#
static void Main(string[] args)
{
    Console = new RealConsole();
...
}
```

Assignment5/HW5/Program.cs
```diff
@@ -0,0 +1,146 @@
...
+            {
+                Console.WriteLine(UnivCalEvent.Display());
+            }
+        }
+
+        public static void CheckForRoomConflicts(List<UniversityCalender> eventList)
+        {
+            foreach (UniversityCalender UnivCalEvent in eventList)
+            {
+                ((Event)UnivCalEvent).RoomsRescourseValidator();
```
Keboo: This works in this application because you only create `Event` classes. However, in reality it would be better to do a type check with the `is` or `as` operator and only call the `RoomsRescourseValidator()` if the cast succeeds.

PR #74, opened 2018-10-31
Assignment5

Nematodes:
The assignment was completed over the course of 3 days, but only committed on the last day because code didn't compile on the previous 2 days.


Assignment5/SchedulePlanner/SchedulePlanner/SchedulePlanner.cs
```diff
@@ -0,0 +1,182 @@
...
+using static System.Environment;
+
+namespace BrianBosAssignmentFive
+{
+    public class SchedulePlanner : IConsole
+    {
+        public static void Main(string[] args)
+        {
+            List<IEvent> eventList = new List<IEvent>();
+            IConsole testSchedulePlanner = new SchedulePlanner();
```
Keboo: This works for running the application, but by creating it inside of the method that it is being used, it prevents the unit tests from being able to provide an alternate implementation. In the [example from class](https://github.com/IntelliTect-Samples/EWU-CSCD371-2018-Fall/blob/master/Lectures/2018.10.23-Interfaces/Interfaces/Program.cs#L18-L20) the `IConsole` implementation was instantiated in the `Main` method, and then _passed_ into another method (a constructor). By separating where it is created from where it is used is the big idea around decoupling your code.

Assignment5/SchedulePlanner/SchedulePlanner/SchedulePlanner.cs
```diff
@@ -0,0 +1,182 @@
...
+                {
+                    testSchedulePlanner.ConsoleOutput($"{NewLine}What kind of event would you like to create?{NewLine}");
+
+                    while (true)
+                    {
+                        testSchedulePlanner.ConsoleOutput($@"1.) Event{NewLine
+                                                            }2.) UniversityCourse{NewLine
+                                                            }3.) Return{NewLine}");
+
+                        userInput = Console.ReadLine();
```
Keboo: In general it is better to not re-use variables for different purposes (in this case as the result for two different menus). For this type of situation, it would be best to have a separate variable for each menu

Assignment5/SchedulePlanner/SchedulePlanner/SchedulePlanner.cs
```diff
@@ -0,0 +1,182 @@
+﻿using System;
+using System.Collections.Generic;
+using static System.Environment;
+
+namespace BrianBosAssignmentFive
+{
+    public class SchedulePlanner : IConsole
```
Keboo: Interesting choice implementing `IConsole` directly on this class. This does technically work and meet the requirements. However, in general I would suggest keeping the [Separation of Concerns](https://deviq.com/separation-of-concerns/) principle in mind. Too much implementation on a single class can result in the class doing more than it should, such as managing the program and performing the console output.

Assignment5/SchedulePlanner/SchedulePlanner/IConsole.cs
```diff
@@ -0,0 +1,7 @@
+﻿namespace BrianBosAssignmentFive
+{
+    public interface IConsole
+    {
+        void ConsoleOutput(string output);
```
Keboo: This is fine, but I would have liked to also see a ReadLine method here so that your program would not need to use `System.Console` directly.

Assignment5/SchedulePlanner/SchedulePlanner.Tests/SchedulePlanner.Tests.cs
```diff
@@ -0,0 +1,789 @@
...
+
+                                      }1.) Create an event{NL
+                                      }2.) List all current events{NL
+                                      }3.) Exit{NL}{NL
+
+                                      }What would you like to do?{NL
+                                      }<<3";
+
+            IntelliTect.TestTools.Console.ConsoleAssert.Expect(expectedOutput, SchedulePlanner.Main);
+        }
```
Keboo: Good unit testing. However, the intent was to be able to replace the `IConsole` implementation used in the main program with a test implementation. Because the main program did not expose a means of doing that (either with a property, or passing it to a method), it left you writing lost of these console based tests. Separating out the console code form that app would have reduced the amount of testing that was needed.

PR #75, opened 2018-10-31
Assignment5 Submission

KennethWhite:
I was looking for some feedback on how to better design my methods to make them testable. It seemed like testing some of the methods in my EventConsoleProgram class would have involved duping the default Console.In and replacing it with a buffer of some sort. I think the idea of the interface around System.Console was to avoid that, any suggestions in regards to this?


Assignment5/Assignment5/Assignment5/EventConsoleProgram.cs
```diff
@@ -0,0 +1,143 @@
...
+                validInput &= int.TryParse(MyConsole.ReadLine(), out int month);
+                MyConsole.WriteLine("Enter the day:");
+                validInput &= int.TryParse(MyConsole.ReadLine(), out int day);
+                MyConsole.WriteLine("Enter the hour:");
+                validInput &= int.TryParse(MyConsole.ReadLine(), out int hour);
+                MyConsole.WriteLine("Enter the minute:");
+                validInput &= int.TryParse(MyConsole.ReadLine(), out int minute);
+                int second = 0;
+                startDate = new DateTime(year, month, day, hour, minute, second);
+            }
```
Keboo: `DateTime` also contains a `TryParse` and `TryParseExact` methods that would make taking in this information from the user simpler since you could accept the information in a single prompt rather than 5.

Assignment5/Assignment5/Assignment5/EventConsoleProgram.cs
```diff
@@ -0,0 +1,143 @@
...
+            catch 
+            {
+                MyConsole.WriteLine("Problem creating date.");
+                startDate = DateTime.Now;
```
Keboo: Typically in `Try*` style methods on a false return you populate the out parameter with a "default" value. In this case something like `default(DateTime)` or `DateTime.MinValue` would have been better.

PR #76, opened 2018-10-31
Assignment5

valdynhunt:


Assignment5/EventApp/src/EventApp/DisplayEvents.cs
```diff
@@ -0,0 +1,41 @@
...
+            string evCollection = "";
+            foreach (Object @object in EventList)
+            {
+                evCollection += Display(@object);
+            }
+
+            return evCollection;
+        }
+
+        public static string Display(Object @object)
```
Keboo: Favor using the keyword `object` over the type name `Object`

Assignment5/EventApp/src/EventApp/Event.cs
```diff
@@ -0,0 +1,154 @@
...
+        {
+            get
+            {
+                return _Name;
+            }
+            set
+            {
+                if (value == null)
+                {
+                    throw new ArgumentNullException(nameof(Name), "Name cannot be null.");
```
Keboo: When throwing exceptions in property setters use `nameof(value)` rather than the name of the property.

Assignment5/EventApp/src/EventApp/MainEvent.cs
```diff
@@ -0,0 +1,93 @@
...
+        {
+            this.Konsole = konsole;
+        }
+
+        public static void Main(string[] args)
+        {
+
+
+            IConsole konsole = new ConcreteMyIConsole();
+            var ME = new MainEvent(konsole);
```
Keboo: Local variables should be pascalCase; `me`.

Assignment5/EventApp/src/EventApp/MockConsole.cs
```diff
@@ -0,0 +1,27 @@
...
+using System.Collections.Generic;
+using System.Text;
+
+namespace EventApp
+{
+    public class MockConsole : IConsole
+    {
+        private int readLineCount = 0;
+        public string ConsoleValue { get; set; }
+        public List<string> InputValue = new List<string>();
```
Keboo: This is a field named like a property.

Assignment5/EventApp/src/EventApp/MockConsole.cs
```diff
@@ -0,0 +1,27 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace EventApp
+{
+    public class MockConsole : IConsole
```
Keboo: Good implementation, however the mock implementation is expected to go into the unit test project not the main project.

Assignment5/EventApp/src/EventApp/UniversityCourse.cs
```diff
@@ -0,0 +1,279 @@
...
+                return _Instructor;
+            }
+            set
+            {
+                if (value == null)
+                {
+                    throw new ArgumentNullException(nameof(Name), "DeptPrefix cannot be null.");
+                }
+
+                if (value.Equals(""))
```
Keboo: There is a `string.IsNullOrEmpty` that you could use to combine these checks into one.

Assignment5/EventApp/src/EventApp/MainEvent.cs
Keboo: This is very close. It is good that you setup the constructor to take is an `IConsole` implementation, however this does not allow your unit test to replace the `ConcreteMyIConsole` implementation with the `MockConsole` implementation. 
What you needed to do was move all of the code in the `Main` method below this to another instance method (like [what we did in class](https://github.com/IntelliTect-Samples/EWU-CSCD371-2018-Fall/blob/master/Lectures/2018.10.23-Interfaces/Interfaces/Program.cs#L20)). This would have let you do something like this in your unit tests.

```
var console = new MockConsole();
//Setup console with whatever it needs to test with.
var mainEvent = new MainEvent(console);
mainEvent.DoStuff();

//Any other checks that are needed
Assert.AreEqual("That's all folks!", console.ConsoleValue);
```

Assignment5/EventApp/test/EventApp.Tests/MainEventTests.cs
```diff
@@ -0,0 +1,157 @@
+using System;
+using System.Diagnostics;
+using EventApp;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace EventApp.Tests
+{
+    [TestClass]
+    public class MainEventTests
```
Keboo: This class name does not match the tests that are in it. A name like this suggests that it will be testing the `MainEvent` class.

PR #80, opened 2018-10-31
Assignment5

Cwhite534:


Assignment5/UniversityCourse/StaticEventLibrary.cs
```diff
@@ -0,0 +1,24 @@
...
+        public static string Display(object @object)
+        {
+            switch (@object)
+            {
+                case UniversityCourse course:
+                    return course.GetSummaryInformation();
+                case Event @event:
+                    return @event.GetSummaryInformation();
+                case null:
+                    throw new ArgumentNullException("Event is null", nameof(@object));
```
Keboo: You have the parameters backwards. The first parameter is suppose to be the name, and the second parameter is the message. It is backwards from the `ArgumentException` so it trips me up all the time too.

Assignment5/UniversityCourse.Tests/IConsoleTester.cs
```diff
@@ -0,0 +1,51 @@
+using System;
+using System.Globalization;
+using System.IO;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace UniversityCourse.Tests
+{
+    [TestClass]
+    public class IConsoleTester
```
Keboo: Strictly speaking you don't write unit tests on interfaces only on the classes that implement the interface.

Assignment5/UniversityCourse.Tests/IConsoleTester.cs
```diff
@@ -0,0 +1,51 @@
...
+        }
+
+        [TestMethod]
+        [ExpectedException(typeof(ArgumentNullException))]
+        public void Test_IConsole_Write_ArgumentNullException()
+        {
+            var myConsole = new TestableConsole();
+            
+            myConsole.WriteLine(null);
+            Assert.Fail();
```
Keboo: You don't need the `Assert.Fail` here. The unit test will fail if the expected exception is not thrown.

Assignment5/UniversityCourse/ConsoleApplication.cs
```diff
@@ -0,0 +1,190 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Globalization;
+using System.Text.RegularExpressions;
+
+namespace UniversityCourse
+{
+    public static class UniversityCourseIo
+    {
+        private static IConsole MyConsole { get;} = new RealConsole();
```
Keboo: This is fine for running the application but it doesn't allow for the unit tests to specify a different implementation. Take a look at [what we did in class](https://github.com/IntelliTect-Samples/EWU-CSCD371-2018-Fall/blob/master/Lectures/2018.10.23-Interfaces/Interfaces/Program.cs#L18-L20). The big idea is to create an `IConsole` instance in your `Main` method then pass it off to another method to use. By passing it as a method parameter rather than storing it in a field, it allows unit tests to pass a `TestableConsole` instance instead. Making the unit testing of this class much simpler.

Assignment5/UniversityCourse/ConsoleApplication.cs
```diff
@@ -0,0 +1,190 @@
...
+            TIME:
+            MyConsole.WriteLine("start time[HH:MM]:");
+            startTime = GetTime();
+            MyConsole.WriteLine("end time[HH:MM]:");
+            endTime = GetTime();
+
+            if (startTime > endTime)
+            {
+                MyConsole.WriteLine("Start time cannot be after end time!");
+                goto TIME;
```
Keboo: In general `goto` statements are typically frowned upon (the exception being inside of switch statements). This type of logic is usually better when converted to a loop statement instead.

Assignment5/UniversityCourse/ConsoleApplication.cs
```diff
@@ -0,0 +1,190 @@
...
+            }
+            
+            return date;
+        }
+        
+        private static DateTime GetTime()
+        {
+            var input = GetNonEmptyString();
+            string[] hrMin;
+            if (new Regex("(?:[01][0-9]|[2][0-4]):(?:[0-5][0-9])").IsMatch(input) && (hrMin = input.Split(':')).Length == 2)
```
Keboo: Good use of Regex

Assignment5/UniversityCourse/ConsoleApplication.cs
```diff
@@ -0,0 +1,190 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Globalization;
+using System.Text.RegularExpressions;
+
+namespace UniversityCourse
+{
+    public static class UniversityCourseIo
```
Keboo: Class name should match the file name

Assignment5/UniversityCourse/UniversityCourse.csproj
```diff
@@ -0,0 +1,8 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>netcoreapp2.0</TargetFramework>
+  </PropertyGroup>
+
+</Project>
```
Keboo: It appears that the entirety of `UniversityCourseIo` is left un-tested. See the comments above, but the intent was to test its method using the `TestableConsole`.

PR #82, opened 2018-10-31
Assignment5

Chester-Southwood:
Assignment5 Submitted


.gitignore
```diff
@@ -1,3 +1,46 @@
+<<<<<<< HEAD
```
Keboo: It looks like you might have a conflict with the .gitignore file that is in master. If you want after class we can address this.

Assignment5/src/Course.cs
```diff
@@ -0,0 +1,174 @@
...
+                {
+                    _Subject = value;
+                }
+                else if(value.Length == 0)
+                {
+                    throw new NotSupportedException("Subject cannot be empty.");           
+                }
+                else
+                {
+                    throw new NullReferenceException("Subject cannot be null.");
```
Keboo: You should not throw NullReferenceExceptions directly. Rather throw an ArgumentNullException. It is also impossible to get to this line because if `value` were null it would throw an exception on the first if statement when it tries to dereference it with `value.Length`

Assignment5/tests/UnitTestMainMethod.cs
```diff
@@ -0,0 +1,13 @@
...
+
+namespace tests
+{
+    [TestClass]
+    public class UnitTestMainMethod
+    {
+        [TestMethod]
+        public void TestMethod1()
+        {
+        }
```
Keboo: The intent here was to have a second implementation of `IConsole` that was testable. This implementation was expected to be passed in place of the `MyConsole` implementation when testing the various methods on `Menu`

Assignment5/src/Menu.cs
```diff
@@ -0,0 +1,66 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Assignment5
+{
+    public class Menu
```
Keboo: This is an un-usual name for your main program class.

Assignment5/src/Menu.cs
```diff
@@ -0,0 +1,66 @@
...
+    {
+        private Event _Events;
+        private static MyConsole myConsole = new MyConsole();
+        public static void Main()
```
Keboo: The intent here was the instantiate `MyConsole` in your `Main` method and pass it as a parameter to the other method (like what [we did in class](https://github.com/IntelliTect-Samples/EWU-CSCD371-2018-Fall/blob/master/Lectures/2018.10.23-Interfaces/Interfaces/Program.cs#L18-L20)). By passing it as a method parameter it allows the unit tests for those same methods to pass an alternate implementation that is more suitable for testing.

Assignment5/src/Menu.cs
```diff
@@ -0,0 +1,66 @@
...
+            int userChoice = -1;
+
+            do{
+                menu();
+                userChoice = getUserChoice();
+                runChoice(eventList, userChoice);
+            }while(userChoice != 3);
+        }
+
+        public static void menu()
```
Keboo: Method names should be camel case.

Assignment5/tests/UnitTestCourse.cs
```diff
@@ -0,0 +1,13 @@
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace tests
+{
+    [TestClass]
+    public class UnitTestCourse
```
Keboo: Is this class needed? You already have `CourseTests`?

Assignment5/src/Course.cs
```diff
@@ -0,0 +1,174 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Assignment5
```
Keboo: Your project name should match your namespace. In this case I would suggest renaming the project to Assignment5. It is fine to have it in a src folder, but I would make sure the project name is more specific.

.gitignore
Chester-Southwood: Will do, I'll be happy to stay after class. It seemed to run okay from my machine before pushing it to github, but I didn't validate after pushing the program. I have little experience with gitignore files, so I appreciate any info about them.

Assignment5/src/Menu.cs
Chester-Southwood: Were you referring userChoice should be userchoice since it's not a method. The method menu is only one word, so I don't know how I would change unless you meant I should have it to be two words such as mainMenu() to signify that it's a method.

Assignment5/src/Menu.cs
Chester-Southwood: Looking back, I agree the error as I have a separate class and method that are essentially called the same word, better naming would be a better improvement on my program.

Assignment5/src/Menu.cs
Keboo: For method names we expected the first letter to be capitalized. So in this case `Menu`

PR #83, opened 2018-10-31
Assignment5

COsborn2:
I would appreciate feedback on how I could refactor my code to make it more testable. Specifically with the methods with a while(true) loop. I was confused about how best to test those methods, or how I could change them to make them easier to test. Thanks!


Assignment5/Assignment5/UniversityCourse.cs
```diff
@@ -0,0 +1,73 @@
...
+        // Validated property
+        public List<char> DaysOfWeek
+        {
+            get => _daysOfWeek;
+
+            set
+            {
+                if (value.Count > 5)
+                {
+                    throw new InvalidDataException("Maximum number of days is 5!");
```
Keboo: Typically we would throw an `ArgumentException` when there is an issue with the incoming argument.

Assignment5/Assignment5/Program.cs
```diff
@@ -0,0 +1,231 @@
...
+            do
+            {
+                selection = PrintMenuAndGetUserSelection();
+                
+                switch (selection)
+                {
+                   case 1:
+                       DateTime startingTime = PromptCreateDateTime();
+                       Event createdEvent = PromptCreateEvent(startingTime);
+                       Console.WriteLine("Event created!");
```
Keboo: The intent of this assignment was to decouple your Program from directly calls to `System.Console`. By doing this it means you have to write long statements for the `ConsoleAssert` class. 
Take a look at [what we did in class](https://github.com/IntelliTect-Samples/EWU-CSCD371-2018-Fall/blob/master/Lectures/2018.10.23-Interfaces/Interfaces/Program.cs#L18-L20). The big idea is to create an `IConsole` instance in your main method as pass it as a method parameter. This then allows your unit test to pass a `TestConsole` when invoking those same methods making unit testing much simpler.

Assignment5/Assignment5.Tests/TestIConsole.cs
```diff
@@ -0,0 +1,55 @@
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace Assignment5.Tests
+{
+    [TestClass]
+    public class TestIConsole
```
Keboo: By definition interfaces do not have functionality so we don't write unit tests against them. Rather we write tests against the classes that implement the interface. This TestConsole class was suppose to be used to test Program.

Assignment5/Assignment5.Tests/TestProgram.cs
```diff
@@ -0,0 +1,221 @@
...
+{
+    // NOTE: only methods without while(true) loops are tested
+    [TestClass]
+    public class TestProgram
+    {
+
+        [TestMethod]
+        public void TestProgramMain_ValidData()
+        {
+            string expected =
```
Keboo: Needing to write long test scripts like this is what can be avoiding by decoupling the code from `System.Console`.

PR #88, opened 2018-11-07
Assignment6 Submission

KennethWhite:


Assignment6/Assignment6/Assignment6.csproj
```diff
@@ -0,0 +1,8 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netcoreapp2.1</TargetFramework>
+    <LangVersion>7.2</LangVersion>
+  </PropertyGroup>
```
KennethWhite: Small note that to make the struct read-only requires C# 7.2

Assignment6/Assignment6.Tests/ScheduleEnumTests.cs
```diff
@@ -0,0 +1,86 @@
...
+            Enum.TryParse<DaysOfWeek>(input, out DaysOfWeek day);
+            Assert.IsTrue(day.HasFlag(expectedFlag));
+
+        }
+
+        [TestMethod]
+        [DataRow("Monday, Tuesday", DaysOfWeek.Monday | DaysOfWeek.Tuesday)]
+        [DataRow("Wednesday, Friday", DaysOfWeek.Wednesday | DaysOfWeek.Friday)]
+        [DataRow("Sunday, Thursday", DaysOfWeek.Sunday | DaysOfWeek.Thursday)]
+        public void ParseDayOfWeek_MultipleDays(string input, System.Enum expectedFlag)
```
breaman: Since this test and the test above do exactly the same thing, you could have combined them both into just 1 test.

Assignment6/Assignment6.Tests/ScheduleEnumTests.cs
```diff
@@ -0,0 +1,86 @@
...
+        {
+            Enum.TryParse<DaysOfWeek>(input, out DaysOfWeek day);
+            Assert.IsTrue(day.HasFlag(expectedFlag));
+        }
+
+        [TestMethod]
+        public void ParseDayOfWeek_AllDays()
```
breaman: Why not make this just another DataRow in the test above?

Assignment6/Assignment6.Tests/StructAssignmentTests.cs
```diff
@@ -0,0 +1,118 @@
...
+        {
+            ValueChanger.ChangeClass(_Cat);
+            Assert.AreEqual(12, _Cat.Age);
+            Assert.AreEqual("Garfield", _Cat.Name);
+        }
+
+        [TestMethod]
+        public void PassStructByReference_ChangeValues_OriginalChanged()
+        {
+            Dog d = new Dog {Name = "Baxter",Age = 99 };
```
breaman: Why is this line here in your test? Looks like d was never used, so this is just unnecessary code.

PR #89, opened 2018-11-07
Assignment6 - Jacob Blazina

blazinaj:
1. Readonly Test case is faked because of compiler error - wasn't sure how to work around that.
2. Can only get struct down to 24 bytes, left the Failing test to illustrate


Assignment6t/Assignment6t/Assignment6tTests/EnumsAndStructsTests.cs
```diff
@@ -0,0 +1,126 @@
...
+    public class EnumsAndStructsTests
+    {
+        [TestMethod]
+        public void DaysOfWeek_Flags_Parse_Correctly()
+        {
+            string enumString = (DaysOfWeek.Monday | DaysOfWeek.Tuesday).ToString();
+
+            Enum.TryParse(enumString, out DaysOfWeek days);
+
+            Assert.AreEqual("Monday, Tuesday", days.ToString());
```
breaman: Should have tested that enum value was actually set correctly instead of testing that ToString printed something out. A better test here would have been Assert.IsTrue(days.HasFlag(DaysOfWeek.Monday | DaysOfWeek.Tuesday)

Assignment6t/Assignment6t/Assignment6tTests/EnumsAndStructsTests.cs
```diff
@@ -0,0 +1,126 @@
...
+        }
+        void MutateInMethodByValue(MyStruct inputStruct)
+        {
+            inputStruct.PropertyOne = inputStruct.PropertyTwo;
+        }
+        void MutateInMethod(MyClass inputClass)
+        {
+            inputClass.PropertyOne = inputClass.PropertyTwo;
+        }
+        void MutateInMethodInstance(MyClass inputClass)
```
breaman: What was supposed to happen in this was that you were supposed to pass inputClass in by ref, create a new instance of MyClass and assign it to inputClass, change a value, and then verify that outside of the method, inputClass had that change.

```
void MutateInMethodInstance(ref MyClass inputClass)
{
   inputClass = new MyClass();
   inputClass.PropertyOne = "some value";
}
```

Assignment6t/Assignment6t/Assignment6tTests/EnumsAndStructsTests.cs
```diff
@@ -0,0 +1,126 @@
...
+
+            Assert.AreEqual(myClass.PropertyTwo, myClass.PropertyOne);
+        }
+
+        [TestMethod]
+        public void Pass_Class_New_Instance_To_Method_Change_Value_Class_Is_Changed()
+        {
+            var myClass = new MyClass(1, 2);
+
+            MutateInMethodInstance(myClass);
```
breaman: myClass should have been passed in by ref here instead.

breaman:
Why were your dll's checked in? Seems like your .gitignore file got all messed up and you had way more files checked in than you should have.

PR #91, opened 2018-11-07
Assignment6  - Raymond Shiner

raymondshiner:


Assignment6/src/Assignment6/MyClass.cs
```diff
@@ -0,0 +1,46 @@
...
+        {
+            myStruct.StructInt = 0;
+        }
+
+        public static void ChangeStructIntToZeroByRef(ref MyStruct myStruct)
+        {
+            myStruct.StructInt = 0;
+        }
+
+        public static void ChangeIStructInterfaceStructIntToZero(ref IStructInterface myStructInterface)
```
breaman: this does not need to have the ref keyword. once you cast the struct to an interface, it is boxed and automatically becomes a reference instead of value type.

Assignment6/src/Assignment6/MyEnums.cs
```diff
@@ -0,0 +1,117 @@
...
+            {
+                Hour = hour;
+                Minute = minute;
+                Second = second;
+            }
+        }
+
+        public struct Schedule
+        {
+            public WeekDays DaysOfWeek;
```
breaman: This is not read-only and your static method that is used to set it proves that. Same goes for your Quarter property.

Assignment6/test/Assignment6.Tests/MyEnums.Tests.cs
```diff
@@ -0,0 +1,51 @@
...
+{
+    [TestClass]
+    public class MyEnumsTests
+    {
+        [TestMethod]
+        public void AddWeekDays_AddMondayToEnums_ReturnsTrue()
+        {
+            MyEnums.Schedule mySchedule = new MyEnums.Schedule();
+
+            MyEnums.AddWeekDays("Monday", ref mySchedule.DaysOfWeek);
```
breaman: You didn't need this method, you could have just use Enum.TryParse to set the value.

Assignment6/test/Assignment6.Tests/MyEnums.Tests.cs
```diff
@@ -0,0 +1,51 @@
...
+
+            Assert.IsTrue(((mySchedule.DaysOfWeek & MyEnums.WeekDays.Monday) != 0));
```
breaman: Assert.AreEqual(MyEnums.WeekDays.Monday, mySchedule.DaysOfWeek) is a little more readable.

Assignment6/test/Assignment6.Tests/MyEnums.Tests.cs
```diff
@@ -0,0 +1,51 @@
...
+        }
+
+        [TestMethod]
+        public void AddWeekDaysm_AddTuesdayAndThursdayToEnums_ReturnsTrue()
+        {
+            MyEnums.Schedule mySchedule = new MyEnums.Schedule();
+
+            MyEnums.AddWeekDays("Tuesday Thursday", ref mySchedule.DaysOfWeek);
+
+            Assert.IsTrue((mySchedule.DaysOfWeek & MyEnums.WeekDays.Tuesday) != 0);
```
breaman: Assert.IsTrue(mySchedule.DaysOfWeek.HasFlag(MyEnums.WeekDays.Monday | MyEnums.WeekDays.Tuesday)) is probably a cleaner way to test this.

PR #93, opened 2018-11-07
Assignment6 - Kris Veselits

kveselits:


Assignment6/UniversityCourse.Tests/SchedulerTests.cs
```diff
@@ -0,0 +1,102 @@
...
+        [TestMethod]
+        public void Test_Valid_Single_Day()
+        {
+            Assert.AreEqual("tuesday", TestSchedule.DayOfWeek.ToString().ToLower());
+        }
+
+        [TestMethod]
+        public void Test_Valid_Multiple_Days()
+        {
+            Assert.AreEqual("tuesday, thursday, friday", TestSchedule2.DayOfWeek.ToString().ToLower());
```
breaman: instead of testing .ToString, a better approach would be Assert.IsTrue(TestSchedule2.DayOfWeek.HasFlag(DaysOfWeek.Tuesday | DaysOfWeek.Thursday | DaysOfWeek.Friday));

Assignment6/UniversityCourse/Scheduler.cs
```diff
@@ -0,0 +1,145 @@
...
+
+        private static DaysOfWeek ParseString(string value)
+        {
+            string[] daysArray = value.Split(' ');
+
+            if (daysArray.Length == 0)
+            {
+                throw new ArgumentException("Bad input");
+            }
+            else if (daysArray.Length == 1)
```
breaman: A nice approach would have been to use the built in Enum.TryParse method that takes a comma separated list of values.

Assignment6/UniversityCourse/Scheduler.cs
```diff
@@ -0,0 +1,145 @@
...
+                case "sunday":
+                    return DaysOfWeek.Sunday;
+                default:
+                    throw new ArgumentException("Invalid input");
+            }
+        }
+
+        private static QuarterOfYear ParseStringQuarter(string value)
+        {
+            switch (value?.ToLower()?.Trim())
```
breaman: Could get rid of this method and just use the Enum.TryParse method instead.

Assignment6/UniversityCourse/Scheduler.cs
```diff
@@ -0,0 +1,145 @@
...
+            {
+                DayOfWeek = ParseString(dayOfWeek);
+                Quarter = ParseStringQuarter(quarterOfYear);
+                StartTime = startTime;
+                Duration = duration;
+            }
+            public DaysOfWeek DayOfWeek { get; }
+            public QuarterOfYear Quarter { get; }
+
+            public Time StartTime
```
breaman: struct was supposed to be readonly, so StartTime and Duration should not have had a set opportunity on the property.

Assignment6/UniversityCourse/Scheduler.cs
```diff
@@ -0,0 +1,145 @@
...
+                set => _Time = value;
+            }
+
+            public TimeSpan Duration
+            {
+                get => _Duration;
+                set => _Duration = value;
+            }
+        }
+        public static void Main(string[] args)
```
breaman: Don't need this method, should be deleted.

Assignment6/UniversityCourse/Scheduler.cs
```diff
@@ -0,0 +1,145 @@
+﻿using System;
+
+namespace UniversityCourse
+{
+    public class Scheduler
+    {
+
+        public static Time _Time;
```
breaman: What are these for? Why are these two properties static on the Scheduler class?

Assignment6/UniversityCourse/Scheduler.cs
```diff
@@ -0,0 +1,145 @@
...
+            Sunday = 1,         //1  0000001
+            Monday = 1 << 1,    //2  0000010
+            Tuesday = 1 << 2,   //4  0000100
+            Wednesday = 1 << 3, //8  0001000
+            Thursday = 1 << 4,  //16 0010000
+            Friday = 1 << 5,    //32 0100000
+            Saturday = 1 << 6,  //64 1000000
+        }
+
+        public enum QuarterOfYear
```
breaman: This should have been of type byte

Assignment6/UniversityCourse/Scheduler.cs
```diff
@@ -0,0 +1,145 @@
...
+                case "winter":
+                    return QuarterOfYear.Winter;
+                default:
+                    throw new ArgumentException(($"Not a valid quarter term: \"{value?.ToString()}\""));
+            }
+        }
+
+        public readonly struct Time
+        {
+            public Time(int hours, int minutes, int seconds)
```
breaman: Hours, Minutes, Seconds should have been of type byte since they don't need full integer values.

Assignment6/UniversityCourse/Scheduler.cs
breaman: Writing these properties "correctly" (not being backed by a static value), it puts this struct at a size of 32, which is larger than the 16 bytes it should have been.

PR #95, opened 2018-11-07
Assignment6

Nematodes:


Assignment6/SchedulePlanner/SchedulePlanner.Tests/ClockTime.Tests.cs
```diff
@@ -0,0 +1,54 @@
...
+    public class ClockTimeTests
+    {
+        private ClockTime TestClockTime;
+
+        [TestMethod]
+        public void Constructor_ValidParameters_FieldsAreSet()
+        {
+            TestClockTime = new ClockTime(5, 13, 21);
+
+            Assert.IsTrue(TestClockTime.Hours == 5);
```
breaman: This isn't bad, but usually when comparing 2 values, you would want to use Assert.AreEqual instead.

PR #96, opened 2018-11-07
Assignment6 Submission

Chester-Southwood:
Submission for Assignment 6, was only able to decrease the size of structure to 24 bytes, resulting in one test failing that checked for size under or equal to 16 bytes.


Assignment6/src/Days.cs
```diff
@@ -0,0 +1,17 @@
+using System;
+
+namespace Assignment6
+{
+        [Flags]
+        public enum Days
```
breaman: This should have been public enum Days : byte

Assignment6/src/Quarters.cs
```diff
@@ -0,0 +1,11 @@
+using System;
+namespace Assignment6
+{
+        public enum Quarters
```
breaman: this should have been public enum Quarters : byte

Assignment6/src/Time.cs
```diff
@@ -0,0 +1,18 @@
+using System;
+
+namespace Assignment6
+{
+    public readonly struct Time
+    {
+        public readonly byte hour;
+        public readonly byte minute;
+        public readonly int second;
```
breaman: this should have been public readonly byte second. you had the right variable type in the constructor.

Assignment6/src/Days.cs
```diff
@@ -0,0 +1,17 @@
...
+        {
+            Monday = 1,
+            Tuesday = 2,
+            Wednesday = 8,
```
breaman: You skipped 4, which means the max value here needed to be 64 instead of 128.

Assignment6/tests/TestOtherStruct.cs
```diff
@@ -0,0 +1,45 @@
...
+        {
+            _TestingSeperateStructure = new SeperateStructRectangle{width = 2, height = 3};
+        }
+
+        [TestMethod]
+        public void UnchangedStructValidTest()
+        {
+            SeperateStructRectangle newStruct = new SeperateStructRectangle{width = 42, height = 42};
+            _TestingSeperateStructure.copyValueToRectangleStructure(newStruct);
+            Assert.AreEqual(_TestingSeperateStructure.height, 3);
```
breaman: Based on your code, this was the wrong value to assert on. You should have asserted on the newStruct since that was the one that got "modified" in your method and we wanted to make sure that when we came back from the method the values were not changed.

Assignment6/tests/TestOtherStruct.cs
```diff
@@ -0,0 +1,45 @@
...
+            _TestingSeperateStructure.copyValueToRectangleStructureReference(ref newStruct);
+            Assert.AreEqual(_TestingSeperateStructure.height, newStruct.height);
+            Assert.AreEqual(_TestingSeperateStructure.width, newStruct.width);
+        }
+        
+        [TestMethod]
+        public void ChangeValueByInterfaceValidTest()
+        {
+            IArea newStruct = _TestingSeperateStructure;
+            newStruct.reset();
```
breaman: This was not quite right. The test works, but what you are testing is not what was asked. What needed to happen here is you should have created another method like your CopyValueToRectangleStructureReference, but used the interface as the input instead. Then you should have noticed that the values in the interface were changed after the call to the method.

Assignment6/src/SeperateStructRectangle.cs
```diff
@@ -0,0 +1,46 @@
+using System;
+namespace Assignment6
+{
+    public class SeperateStructRectangle: IArea
+    {
+        public double width
```
breaman: properties and methods should be PascalCase.

Assignment6/tests/TestOtherClass.cs
```diff
@@ -0,0 +1,36 @@
...
+            _TestingSeperateClass = new SeperateClassSquare(2,3);
+        }
+
+        [TestMethod]
+        public void UnchangedClassValidTest()
+        {
+            SeperateClassSquare newStruct = new SeperateClassSquare(42,42);
+            _TestingSeperateClass.copyValueToSquareStructure(newStruct);
+            Assert.AreEqual(_TestingSeperateClass.height, newStruct.height);
+            Assert.AreEqual(_TestingSeperateClass.width, _TestingSeperateClass.width);
```
breaman: This assert is bogus

Assignment6/tests/TestOtherClass.cs
```diff
@@ -0,0 +1,36 @@
...
+        }
+
+        [TestMethod]
+        public void ChangedClassByReferenceValidTest()
+        {
+            SeperateClassSquare newStruct = new SeperateClassSquare(42,42);
+            _TestingSeperateClass.copyValueToSquareStructureReference(ref newStruct);
+            Assert.AreEqual(_TestingSeperateClass.height, newStruct.height);
+            Assert.AreEqual(_TestingSeperateClass.width, _TestingSeperateClass.width);
```
breaman: This assert is bogus.

Assignment6/tests/TestOtherClass.cs
```diff
@@ -0,0 +1,36 @@
...
+            _TestingSeperateClass.copyValueToSquareStructure(newStruct);
+            Assert.AreEqual(_TestingSeperateClass.height, newStruct.height);
+            Assert.AreEqual(_TestingSeperateClass.width, _TestingSeperateClass.width);
+        }
+
+        [TestMethod]
+        public void ChangedClassByReferenceValidTest()
+        {
+            SeperateClassSquare newStruct = new SeperateClassSquare(42,42);
+            _TestingSeperateClass.copyValueToSquareStructureReference(ref newStruct);
```
breaman: What should have been tested here is that a whole new object with new values was assigned to "newStruct" (which is a bad name since this isn't a struct) and then when newStruct came back from the method call, the new values (and the new instance) were modified.

PR #97, opened 2018-11-07
Assignment6

Cwhite534:


breaman:
Very nice work on the assignment. Unit tests were thorough and everything that was asked for in the assignment was completed.

PR #98, opened 2018-11-07
Assignment6

joeriddles:
See Assignment6 README for new code changes.


Assignment6/src/Vehicle/VehicleUtils.cs
```diff
@@ -0,0 +1,38 @@
...
+
+		public static Car UpgradeHorsePower(Car car)
+		{
+			car.HorsePower += 100;
+			return car;
+		}
+
+		public static Car NewCarWithSpoiler(Car car)
+		{
+			Car newCar = car;
```
breaman: this should have had the variable passed in by ref, then used the statement car = new Car(). this just assigned a new instance to the same reference as the current one and then changed the value of the new reference pointer.

Assignment6/src/Schedule/IEvent.cs
```diff
@@ -0,0 +1,49 @@
...
+		string StartTime { get; set; }
+		string GetSummaryInformation();
+	}
+
+	[Flags]
+	public enum Days
+	{
+		None = 0,
+		Monday = 1,
+		Tuesday = 1 << 2,
```
breaman: This should have started off with just 1 << 1, you didn't need to shift 2 bits right off the bat.

Assignment6/src/Schedule/IEvent.cs
```diff
@@ -0,0 +1,49 @@
...
+{
+	public interface IEvent
+	{
+		string Description { get; set; }
+		string StartTime { get; set; }
+		string GetSummaryInformation();
+	}
+
+	[Flags]
+	public enum Days
```
breaman: This should have been public enum Days : byte

Assignment6/src/Schedule/IEvent.cs
```diff
@@ -0,0 +1,49 @@
...
+		Tuesday = 1 << 2,
+		Wednesday = 1 << 3,
+		Thursday = 1 << 4,
+		Friday = 1 << 5,
+		Saturday = 1 << 6,
+		Sunday = 1 << 7
+	}
+
+	[Flags]
+	public enum Quarters
```
breaman: This should have been public enum Quarters : byte

Assignment6/src/Schedule/IEvent.cs
```diff
@@ -0,0 +1,49 @@
...
+	{
+		None = 0,
+		Spring = 1,
+		Summer = 1 << 2,
```
breaman: Should have only bit shifted this 1 place first.

Assignment6/src/Schedule/IEvent.cs
```diff
@@ -0,0 +1,49 @@
...
+		Fall = 1 << 3,
+		Winter = 1 << 4
+	}
+
+	public readonly struct Time
+	{
+		public int Hours { get; }
```
breaman: Since we are only storing values 0 - 23 and 0-59, these all could have been byte instead of int.

Assignment6/src/Schedule/EventExtensions.cs
```diff
@@ -0,0 +1,43 @@
+﻿namespace Schedule
+{
+	public static class EventExtensions
+	{
+		public static Days ParseDaysFromString(this string daysString)
+		{
+			Days days = Days.None;
+
+			if (daysString.Contains("Monday"))
```
breaman: Instead of a all these if statements, you could have just done
```
Enum.TryParse(daysString, out days);
```

Assignment6/src/Schedule/EventExtensions.cs
```diff
@@ -0,0 +1,43 @@
...
+				days |= Days.Saturday;
+			if (daysString.Contains("Sunday"))
+				days |= Days.Sunday;
+
+			return days;
+		}
+
+		public static Quarters ParseQuartersFromString(this string quartersString)
+		{
+			Quarters quarters = Quarters.None;
```
breaman: Should have used Enum.TryParse here as well instead of all the if statements.

PR #113, opened 2018-11-16
Assignment8

JeffHowes:


Assignment8/HW8-WPF/MainWindow.xaml.cs
```diff
@@ -0,0 +1,121 @@
...
+            return idateTime.DateTime(); 
+        }
+
+        private void Timer_Click(object sender, EventArgs e)
+        {
+            
+            DateTime now = idateTime.DateTime();
+            TimeSpan elapseTime = (now - _startTickTime);
+            Timer.Content = elapseTime.Hours + ":" + elapseTime.Minutes 
+                + ":" + elapseTime.Seconds + "." + elapseTime.Milliseconds;
```
Keboo: This could be simplified with a [format string](https://docs.microsoft.com/en-us/dotnet/standard/base-types/standard-timespan-format-strings).

Assignment8/HW8-WPF/MainWindow.xaml.cs
```diff
@@ -0,0 +1,121 @@
...
+
+        public void ResetButton_OnClick(object sender, EventArgs e)
+        {
+            _stoppedTickTime = _startTickTime;
+            Timer.Content = "00:00:00.000";
+        }
+
+        public void SaveButton_OnClick(object sender, EventArgs e)
+        {            
+            StopButton_OnClick(this, null);
```
Keboo: Typically you don't want to invoke an event handler directly like this. If the handler were to expect the sender to be the button or the EventArgs to be populated this would break. If you need to share functionality it to better to move the code out of the event handler method into a new method with no parameters that you call in both places.

Assignment8/HW8-WPF/MainWindow.xaml.cs
```diff
@@ -0,0 +1,121 @@
...
+        public void SaveButton_OnClick(object sender, EventArgs e)
+        {            
+            StopButton_OnClick(this, null);            
+             TimeSpan elapsedTime = _stoppedTickTime - _startTickTime;            
+            TasksListBox.Items.Add(new CompletedTaskItem() { Title = TaskTitleTextBox.Text,  ElapsedTime = elapsedTime });
+        }
+
+        private void DeleteButton_Click(object sender, RoutedEventArgs e)
+        {
+            if (TasksListBox.SelectedItem == null)
```
Keboo: I would invert this if statement. Typically you don't want an empty if block with just an else.
```C#
if (TasksListBox.SelectedItem != null)
{
    TasksListBox.Items.RemoveAt
        (TasksListBox.Items.IndexOf(TasksListBox.SelectedItem));
}
```

Assignment8/HW8-WPF/TestableDateTime.cs
```diff
@@ -0,0 +1,23 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace HW8_WPF
+{
+    public class TestableDateTime : IDateTime
```
Keboo: This is a great test implementation. But you test implementation is expected to be inside of the unit test project not inside of the main WPF project. Take a look at the sample assignment 8 I put up if you want to see an example.

Assignment8/HW8-WPF/TimeManager.cs
```diff
@@ -0,0 +1,54 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace HW8_WPF
+{
+    public class TimeManager
```
Keboo: I think the key thing you are missing here, is that the TimeManager's constructor should take in an `IDateTime` and use this to get the current time for the Start/Stop methods rather than taking in a `DateTime` directly.

PR #114, opened 2018-11-16
Assignment8 Submission

COsborn2:
Everything that WPF gave me is included in this submission. Im assuming some of these aren't needed for the project, and are just compiled binaries, but I wasn't sure, so I left them in.


Assignment8/Assignment8/MainWindow.xaml
```diff
@@ -0,0 +1,60 @@
...
+        xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
+        xmlns:local="clr-namespace:Assignment8"
+        mc:Ignorable="d"
+        Title="MainWindow" Height="450" Width="600"
+        MinWidth="450"
+        MinHeight="600">
+    
+    
+    <Border Padding="10">
+        <Grid x:Name="MainGrid">
```
Keboo: You don't need to set the `x:Name` on elements that you don't need to reference in the code behind (C# file).

Assignment8/Assignment8/MainWindow.xaml.cs
```diff
@@ -0,0 +1,58 @@
...
+
+        public void OnStopClicked(object sender, EventArgs args)
+        {
+            // do something
+            ((TimeManager)DataContext).TimerStop(DescriptionText.Text);
+        }
+
+        public void CreateNewListBoxEntry(TimeEvent passedEvent)
+        {
+            TimeEvent eventToAdd = passedEvent;
```
Keboo: You really don't need this local variable. Just pass `passedEvent` directly to the `TimeList.Items.Add`

Assignment8/Assignment8/TimeEvent.cs
```diff
@@ -0,0 +1,26 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Assignment8
+{
+    public class TimeEvent
+    {
+        private string _timeElapsed { get; set; }
```
Keboo: This property is named like a field.

Assignment8/Assignment8/TimeManager.cs
```diff
@@ -0,0 +1,125 @@
...
+using System.Diagnostics;
+using System.Linq;
+using System.Runtime.CompilerServices;
+using System.Text;
+using System.Threading.Tasks;
+using System.Windows.Threading;
+
+namespace Assignment8
+{
+    public class TimeManager : IDateTime, INotifyPropertyChanged
```
Keboo: This class is doing too much. The intent was to decouple it from calling `DateTime.Now`. To do this you need to have a separate class implement `IDateTime`. This class then should take in an `IDateTime` in its constructor and use it whenever it needs to get the current time.

Assignment8/Assignment8/TimeManager.cs
```diff
@@ -0,0 +1,125 @@
...
+            {
+                if (_currentTime != value)
+                {
+                    _currentTime = value;
+                    OnPropertyChanged();
+                }
+            }
+        }
+
+        public TimeManager(Action<TimeEvent> stopEvent)
```
Keboo: The `TimeManager` class was suppose to have an [event](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/events/) (see chapter 14) not just a callback delegate.

Assignment8/Assignment8/TimeManager.cs
```diff
@@ -0,0 +1,125 @@
...
+                _elapsedTimer.Start();
+                _isTimerRunning = true;
+            }
+        }
+
+        public void OnTickElapsed(object sender, EventArgs e)
+        {
+            DateTime now = GetCurrentTime();
+            TimeSpan elapsedTime = now - _startTime;
+            CurrentTime = elapsedTime.ToString(@"hh':'mm':'ss':'ff");
```
Keboo: Rather than trying to keep track of the `CurrentTime` inside of the timer's Tick event it is better to get the current time when the timer stops and compare against the start time. For simplistic time entry where milliseconds don't matter this will "work" but for any application where accuracy is important this method will quickly fall apart.

Assignment8/Assignment8/TimeManager.cs
```diff
@@ -0,0 +1,125 @@
...
+using System.Text;
+using System.Threading.Tasks;
+using System.Windows.Threading;
+
+namespace Assignment8
+{
+    public class TimeManager : IDateTime, INotifyPropertyChanged
+    {
+        private DispatcherTimer _currentTimer;
+        private DispatcherTimer _elapsedTimer;
```
Keboo: Typically DispatcherTimers are treated like UI elements and kept inside of the Window class rather than in the class being unit tested (as it makes the class more difficult to test). Take a look at the sample solution for Assignment 8 I put if you want to see an example of this.

Assignment8/Assignment8/TimeManager.cs
```diff
@@ -0,0 +1,125 @@
...
+        private DateTime _startTime;
+        private bool _isTimerRunning;
+
+        private string _currentTime;
+        public string CurrentTime
```
Keboo: Good implementation of a bindable property.

Assignment8/Assignment8/TimeManager.cs
```diff
@@ -0,0 +1,125 @@
...
+        public void TimerStart()
+        {
+            if(!_isTimerRunning)
+            {
+                // stop clock timer
+                _currentTimer.Stop();
+
+                _elapsedTimer = new DispatcherTimer();
+                _elapsedTimer.Interval = TimeSpan.FromMilliseconds(10);
+                _elapsedTimer.Tick += OnTickElapsed;
```
Keboo: This is throwing away the old `DispatcherTimer` instance and re-creating a new one each time. It would be better to instantiate this once in the constructor and simply call `Start()` and `Stop()` on it.

Assignment8/Assignment8/TimeManager.cs
```diff
@@ -0,0 +1,125 @@
...
+                StopEvent(toSend);
+
+                //start clock
+                StartClock();
+            }
+        }
+
+        public Action<TimeEvent> StopEvent { get; set; }
+
+        public bool IsClockEnabled()
```
Keboo: This could be simplified to just be:
```C#
public bool IsClockEnabled() => _currentTimer?.IsEnabled == true;
```

Assignment8/Assignment8/TimeManager.cs
```diff
@@ -0,0 +1,125 @@
...
+                TimeEvent toSend = new TimeEvent(CurrentTime, eventDescription);
+
+                StopEvent(toSend);
+
+                //start clock
+                StartClock();
+            }
+        }
+
+        public Action<TimeEvent> StopEvent { get; set; }
```
Keboo: This should have been an `event` not just a delegate.

Assignment8/Assignment8.Tests/TimeManagerTests.cs
```diff
@@ -0,0 +1,126 @@
...
+            SetStringOnCompletion = "success";
+        }
+
+        public void MockPassedEvent(TimeEvent toUse)
+        {
+            SetTimeEventOnTimerStop = toUse;
+        }
+    }
+
+    class TestableDateTime : IDateTime
```
Keboo: This class should be in its own file.

Assignment8/Assignment8.Tests/TimeManagerTests.cs
```diff
@@ -0,0 +1,126 @@
...
+        }
+
+        [TestMethod]
+        public void TestClockStartAtRightTime_CurrentSystemTimeEqualToTimeManagerTime()
+        {
+            DateTime beforeTicking = DateTime.Now;
+
+            TimeManagerForTests = new TimeManager(PassMethod);
+
+            Assert.AreEqual(beforeTicking.ToString(@"hh \: mm :\ ss"), TimeManagerForTests.CurrentTime);
```
Keboo: Though this test will pass most of the time, it is not guaranteed to pass all of the time. Imagine if `beforeTicking` were to get a time like: 0:0:0.9999999999 and it takes more than a fraction of a second to instantiate the time manager and the `TimeManager.CurrentTime` gets the value 0:0:1.000000000. 
This is the reason for needing to pass an instance of `IDateTime` into the constructor of the `TimeManager` class. In this setup you can make the `TimeManager` behave in a consistent manner.

Assignment8/Assignment8.Tests/TimeManagerTests.cs
```diff
@@ -0,0 +1,126 @@
...
+        }
+
+        [TestMethod]
+        public void TestClockIsTicking_ClockIsEnabledReturnsTrue()
+        {
+            DateTime beforeTicking = DateTime.Now;
```
Keboo: Variable un-used.

Assignment8/Assignment8.Tests/TimeManagerTests.cs
```diff
@@ -0,0 +1,126 @@
...
+
+            TimeManagerForTests = new TimeManager(PassMethod);
+
+            Assert.IsTrue(TimeManagerForTests.IsClockEnabled());
+        }
+
+        [TestMethod]
+        public void TestTimerIsTicking_TimerIsEnabledReturnsTrue()
+        {
+            DateTime beforeTicking = DateTime.Now;
```
Keboo: Variable un-used.

Assignment8/Assignment8.Tests/TimeManagerTests.cs
```diff
@@ -0,0 +1,126 @@
...
+
+            TimeManagerForTests.TimerStart();
+
+            Assert.IsTrue(TimeManagerForTests.IsTimerEnabled());
+        }
+
+        [TestMethod]
+        public void TestTimerNotTicking_TimerNotTickingBeforeStarted()
+        {
+            DateTime beforeTicking = DateTime.Now;
```
Keboo: Variable un-used.

PR #117, opened 2018-11-16
Assignment8

valdynhunt:


Assignment8/TimeTracker/TimeTracker/TimeManager.cs
```diff
@@ -0,0 +1,56 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace WpfApp1
+{
+    public class TimeManager : IDateTime
```
Keboo: You wanted to have a separate class implement `IDateTime` rather than making the `TimeManager` class do it. This is what would allow you in your unit tests to properly test its methods by being able to control what the "current" time is.

Assignment8/TimeTracker/TimeTrackerTest/TestableDateTime.cs
```diff
@@ -0,0 +1,32 @@
+﻿using System;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using TimeTracker;
+
+namespace TimeTrackerTest
+{
+
+    public class TestableDateTime : IDateTime
```
Keboo: This is a good testable implementation!

PR #119, opened 2018-11-16
Assignment8

toha87:


Assignment8/Assignment8/MainWindow.xaml
```diff
@@ -0,0 +1,60 @@
+﻿<Window x:Class="Assignment8.MainWindow"
+        xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
+        xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+        xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
+        xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
+        xmlns:local="clr-namespace:Assignment8"
+        mc:Ignorable="d"
+        Title="MainWindow" Height="450" Width="1000">
+    <Grid x:Name="MyGrid" Margin="40,10">
```
Keboo: You don't need to set the `x:Name`/`Name` on elements that you don't need to reference in the code behind (C# file).

Assignment8/Assignment8/MainWindow.xaml.cs
```diff
@@ -0,0 +1,82 @@
...
+
+        private void ClockTime_Tick(object sender, EventArgs e)
+        {
+            IDateTime dateTime = new TimeNow();
+            labelClock.Content = dateTime.DateTime();  
+        }
+
+        private void Timer_Tick(object sender, EventArgs e)
+        { 
+            currentTimer = Convert.ToString(DateTime.Now - start);
```
Keboo: Though this "works" it is very inaccurate. For an app like this where it can be off by a little bit it is fine, but in an application where accuracy matters you should calculate the stop time when you click the Stop button.

Assignment8/Assignment8/MainWindow.xaml.cs
```diff
@@ -0,0 +1,82 @@
...
+        private DateTime start;
+        private string currentTimer;
+
+        public MainWindow()
+        {
+            InitializeComponent();
+
+            _clock = new DispatcherTimer
+            {
+                Interval = new TimeSpan(0, 0, 1)
```
Keboo: For `TimeSpan` there is a static method that is a little easier to read. `TimeSpan.FromSeconds(1)`.

Assignment8/Assignment8/MainWindow.xaml.cs
```diff
@@ -0,0 +1,82 @@
...
+            _timer.Start();
+            start = time.LabelTimeCurrent();
+            _timer.Tick += Timer_Tick;
+        }
+
+        private void ClickMeButton_OnClickStopTimer(object sender, RoutedEventArgs e)
+        {
+            _timer.Stop();
+
+            ListBox.Items.Add(new TimeManager()
```
Keboo: It seems like you are using the TimeManager class for several different things. Perhaps there should be a separate class for this.

Assignment8/Assignment8/MainWindow.xaml.cs
```diff
@@ -0,0 +1,82 @@
...
+        private void ClickMeButton_OnClickStartTimer(object sender, EventArgs e)
+        {
+            TimeManager time = new TimeManager();
+            _timer.Start();
+            start = time.LabelTimeCurrent();
+            _timer.Tick += Timer_Tick;
+        }
+
+        private void ClickMeButton_OnClickStopTimer(object sender, RoutedEventArgs e)
+        {
```
Keboo: Because there are no checks in this method it is possible to click the Stop button multiple times and add several items to the list box. It would be good to check if there was a timer started.

Assignment8/Assignment8/MainWindow.xaml.cs
```diff
@@ -0,0 +1,82 @@
...
+            _timer.Stop();
+
+            ListBox.Items.Add(new TimeManager()
+            {
+                EventName = EventTextBox.Text.ToString(),
```
Keboo: `EventTextBox.Text` returns a `string` so the `ToString()` call is not needed here.

Assignment8/Assignment8/TimeManager.cs
```diff
@@ -0,0 +1,16 @@
+﻿
+using System;
+
+namespace Assignment8UnitTest
+{
+    public class TimeManager
```
Keboo: The `TimeManager` class was suppose to have an [event](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/events/) (see chapter 14).

PR #120, opened 2018-11-16
Assignment8 Submission - Raymond Shiner

raymondshiner:


raymondshiner:
I just Realized that on line 92 of Mainwindow.Xaml.cs, i assign StartTime, but it never gets used. This is because i forgot to add to the string on lines 98 and 99 of the same file(when the task completes) this code 

97 - DateTime now = TimeManager.Now();
98 - DateTime interval = now - StartTime;
99 (Append to End of String) - " - Task was completed in interval seconds".

Basically i meant to in that method utilize startTime to also list the interval. Could've done it in seconds just forgot cuz i was frantically trying to get my TimeManager class to work. Don't expect a grade change or anything just trying to explain my intention for what ended up being useless code.

Assignment8/Assignment8/TimeManager.cs
```diff
@@ -0,0 +1,31 @@
...
+    {
+        public Action StartStopButtonClick { get; set; }
+        public Action DeleteButtonClick { get; set; }
+
+        public void RaiseEventStartStop(object sender, RoutedEventArgs routedEventArgs)
+        {
+            StartStopButtonClick?.Invoke();
+        }
+
+        public void RaiseEventDelete(object sender, RoutedEventArgs e)
```
Keboo: The delete functionality didn't need to go through the TimeManager and would have been simpler to do in the MainWindow directly.

Assignment8/Assignment8/TimeManager.cs
```diff
@@ -0,0 +1,31 @@
...
+using System.Windows;
+
+namespace Assignment8
+{
+    public class TimeManager : IDateTime
+    {
+        public Action StartStopButtonClick { get; set; }
+        public Action DeleteButtonClick { get; set; }
+
+        public void RaiseEventStartStop(object sender, RoutedEventArgs routedEventArgs)
```
Keboo: I would have expected this method to do things like keep track of the timer state (basically what your `ThisClickIsStart` boolean in MainWindow is doing).

Assignment8/Assignment8/MainWindow.xaml.cs
```diff
@@ -0,0 +1,113 @@
...
+            _timer.Start();
+            TaskRunningMessage.Content = "Press Start to Start a Task";
+            ThisClickIsStart = false;
+
+            var manager = new TimeManager();
+            manager.StartStopButtonClick += StartStop_OnClick_SetCurrentClick;
+            manager.StartStopButtonClick += Stop_OnClick_DisplayTaskFinished;
+            manager.StartStopButtonClick += Stop_OnClick_EndTask;
+            manager.StartStopButtonClick += Start_OnClick_DisplayTaskRunning;
+            manager.StartStopButtonClick += Start_OnClick_StartTask;
```
Keboo: Though this works, it does leave the code very disjointed and difficult to follow. Typically you only have a single event handler in a given class.

PR #121, opened 2018-11-16
Assignment8

Chester-Southwood:
I misused my time during the weekend focusing on trying the extra credit of MVVM instead of the assignment in general. 
I did not unit test as is required nor did properly invoke event. The way I had my code coupled didn't allow me to unit test easily as was warned in the documentation.


Assignment8/Assignment8/MainWindow.xaml
```diff
@@ -0,0 +1,117 @@
+﻿<Window x:Class="Assignment8.MainWindow"
+        xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
+        xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+        xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
+        xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
+        xmlns:local="clr-namespace:Assignment8"
+        mc:Ignorable="d"
+        Title="MainWindow" Height="450" Width="800" MinWidth="350">
+    <!--Window start-->
+    <Grid x:Name="MyGrid">
```
Keboo: You don't need to set the `x:Name`/`Name` on elements that you don't need to reference in the code behind (C# file).

Assignment8/Assignment8/MainWindow.xaml.cs
```diff
@@ -0,0 +1,138 @@
...
+
+namespace Assignment8
+{
+    /// <summary>
+    /// Interaction logic for MainWindow.xaml
+    /// </summary>
+    public partial class MainWindow : Window
+    {
+        private DispatcherTimer _Timer;
+        private DateTime _LastTickTime;
```
Keboo: Field appears to be un-used

Assignment8/Assignment8/MainWindow.xaml.cs
```diff
@@ -0,0 +1,138 @@
...
+        private DateTime _TimeStarted;
+        private Boolean _IsTimerOn = false;
```
Keboo: Favor using keywords instead of class name.

Assignment8/Assignment8/MainWindow.xaml.cs
```diff
@@ -0,0 +1,138 @@
...
+            //}
+        }
+
+        private void EndTimer_Click(object sender, RoutedEventArgs e)
+        {
+            if (TitleTextBox.Text != null && TitleTextBox.Text.Length > 0)
+            {
+                string time = CurrentTimeBlock.Text.ToString();
+                //TimeSpan t = DateTime.Now - _TimeStarted;
+                _Timer = new DispatcherTimer();
```
Keboo: This is throwing away the old `DispatcherTimer` instance and re-creating a new one each time. It would be better to instantiate this once in the constructor and simply call `Start()` and `Stop()` on it.
This opens up a bug where you timer keeps ticking after you click stop because the old timer is still running.

Assignment8/Assignment8/MainWindow.xaml.cs
```diff
@@ -0,0 +1,138 @@
...
+        {
+            if (!_IsTimerOn)
+            {
+                MyStartTimer();
+                _IsTimerOn = true;
+            }
+        }
+
+        private void PauseTimer_Click(object sender, RoutedEventArgs e)
+        {
```
Keboo: It is nice being able to Pause, but with no way to resume the timer. This would have been better named "StopTimer"

Assignment8/Assignment8/MainWindow.xaml.cs
```diff
@@ -0,0 +1,138 @@
...
+        }
+
+
+        private void ResetTimer_Click(object sender, RoutedEventArgs e)
+        {
+            /*if (_IsTimerOn)
+            {*/
+                MyStartTimer();
+                _Timer.Stop();
+                _IsTimerOn = false;
```
Keboo: Rather than using `_IsTimerOn` you could simply use `_Timer.IsEnabled` which will given you the current state of the timer

Assignment8/Assignment8/MainWindow.xaml.cs
```diff
@@ -0,0 +1,138 @@
...
+                ResetTextBlock(TitleTextBox);
+            //}
+        }
+
+        private void EndTimer_Click(object sender, RoutedEventArgs e)
+        {
+            if (TitleTextBox.Text != null && TitleTextBox.Text.Length > 0)
+            {
+                string time = CurrentTimeBlock.Text.ToString();
```
Keboo: `CurrentTimeBlock.Text` returns a string so the `ToString()` call is not needed.

Assignment8/Assignment8/IDateTime.cs
```diff
@@ -0,0 +1,13 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Assignment8
+{
+    interface IDateTime
+    {
+        DateTime getDateTime();
```
Keboo: Casing. Method should start with a capital letter. `GetDateTime()`

PR #123, opened 2018-11-16
Assignment8

joeriddles:


Assignment8/src/DateTime/TimeManager.cs
```diff
@@ -0,0 +1,27 @@
+﻿namespace DateTime
+{
+	public class TimeManager
```
Keboo: This is perfect. The only thing it is missing is an event that gets raised in the EndTimer method to indicate that an entry should be created.

Assignment8/src/Timer/Timer/App.xaml
```diff
@@ -0,0 +1,16 @@
...
+             xmlns:local="clr-namespace:Timer"
+             StartupUri="MainWindow.xaml">
+    <Application.Resources>
+        <ResourceDictionary>
+            <ResourceDictionary.MergedDictionaries>
+                <ResourceDictionary Source="pack://application:,,,/MaterialDesignThemes.Wpf;component/Themes/MaterialDesignTheme.Light.xaml" />
+                <ResourceDictionary Source="pack://application:,,,/MaterialDesignThemes.Wpf;component/Themes/MaterialDesignTheme.Defaults.xaml" />
+                <ResourceDictionary Source="pack://application:,,,/MaterialDesignColors;component/Themes/Recommended/Primary/MaterialDesignColor.DeepPurple.xaml" />
+                <ResourceDictionary Source="pack://application:,,,/MaterialDesignColors;component/Themes/Recommended/Accent/MaterialDesignColor.Lime.xaml" />
+            </ResourceDictionary.MergedDictionaries>
```
Keboo: Great job including MDIX!

Assignment8/src/Timer/Timer/MainWindow.xaml
```diff
@@ -0,0 +1,40 @@
...
+
+        <materialDesign:Card Padding="32" Margin="16" Grid.ColumnSpan="2">
+            <TextBlock x:Name="MyClock" TextAlignment="Center" FontSize="72"/>
+        </materialDesign:Card>
+
+        <Button x:Name="StartStopTimer" Click="StartStopTimer_OnClick" Content="Start Timer" Background="Green" BorderBrush="Green" Grid.Row="1" Grid.Column="0"/>
+        <Button x:Name="Delete" Click="Delete_OnClick" Content="Delete Entry" Background="Red" BorderBrush="Red" Grid.Row="1" Grid.Column="1"/>
+
+        <ListBox  x:Name="TimeEntries" Grid.Row="2" Grid.Column="0" Grid.ColumnSpan="2"/>
+        <materialDesign:Card x:Name="LoadEntries" Grid.Row="2" Grid.Column="0" Grid.ColumnSpan="2"></materialDesign:Card>
```
Keboo: Not sure why this card is here. It doesn't contain any contents so it gets a height of 0 and is effectively invisible.  It is technically on top of your list box so if you were to put anything in it, it would show up on top.

Assignment8/src/Timer/Timer/MainWindow.xaml.cs
```diff
@@ -0,0 +1,88 @@
...
+		private string _endTime;
+		private int _entryCounter = 1;
+
+		public MainWindow()
+		{
+			InitializeComponent();
+			Clock clock = new Clock();
+			clock.Add((sender, e) => { MyClock.Text = clock.CurrentTime; });
+			IDateTime dateTimeInterface = new DateTime.DateTime();
+			_manager = new TimeManager(dateTimeInterface);
```
Keboo: Great!

Assignment8/src/Timer/Timer/MainWindow.xaml.cs
```diff
@@ -0,0 +1,88 @@
...
+			{
+				_startTime = _manager.StartTimer();
+				StartStopTimer.Content = "Stop Timer";
+				StartStopTimer.Background = new SolidColorBrush(Colors.OrangeRed);
+				StartStopTimer.BorderBrush = new SolidColorBrush(Colors.OrangeRed);
+			}
+			else
+			{
+				_endTime = _manager.EndTimer();
+				TimeEntries.Items.Add(CreateEntry());
```
Keboo: Rather than doing this here. The expectation was that the TimeManager would raise an event that would trigger the adding of a new entry.

Assignment8/src/Timer/Timer/MainWindow.xaml.cs
```diff
@@ -0,0 +1,88 @@
...
+		}
+
+		private UIElement CreateEntry()
+		{
+			StackPanel vStack = new StackPanel();
+			vStack.Children.Add(new TextBlock
+				{Text = $"Entry #{_entryCounter} Start: {_startTime} for {_endTime} seconds.", FontSize = 16});
+
+			StackPanel hStack = new StackPanel {Orientation = Orientation.Horizontal};
+			hStack.Children.Add(new Label {Content = "Description:"});
```
Keboo: In WPF you want to use a `TextBlock` to display simple text. You only need to use a `Label` when you need to [set an access key](https://docs.microsoft.com/en-us/dotnet/api/system.windows.controls.label.target?view=netframework-4.7.2) to move the keyboard focus (something we did not cover in class).

Assignment8/test/DateTime.Tests/DateTimeTests.cs
```diff
@@ -0,0 +1,22 @@
...
+namespace DateTime.Tests
+{
+	[TestClass]
+	public class DateTimeTests
+	{
+		[TestMethod]
+		public void DateTime_NewDateTime_Success()
+		{
+			TimeManagerTests.DateTimeTest datetime = new TimeManagerTests.DateTimeTest();
+			Assert.IsNotNull(datetime);
```
Keboo: This test does not add much value. It is not possible to call the constructor and get back a `null` instance

Assignment8/test/DateTime.Tests/TimeManagerTests.cs
```diff
@@ -0,0 +1,56 @@
...
+		}
+
+		[TestMethod]
+		public void TimeManager_New_Success()
+		{
+			IDateTime testDateTimeInterface = new DateTimeTest();
+			TimeManager manager = new TimeManager(testDateTimeInterface);
+			Assert.IsNotNull(manager);
+			Assert.AreEqual(typeof(DateTimeTest).ToString(), manager.DateTime.ToString());
+		}
```
Keboo: This test confuses me a little. It is not possible to call the constructor and get back a `null` instance. I am not sure what the point is of testing the base implementation of `ToString()`

PR #132, opened 2018-12-04
Assignment9

toha87:


Assignment9/Assignment9.Tests/Assignment9.UnitTests.cs
```diff
@@ -0,0 +1,80 @@
...
+        public void ReturnsAnIEnumerableTOfOriginalItemsInRandomOrder()
+        {
+            IEnumerable<Inventor> list = PatentData.Inventors.ToList();
+
+
+            for (int i = 0; i < 3; i++)
+            {
+                IEnumerable<Inventor> listToTest = PatentData.Inventors.Randomize().ToList();
+
+                Assert.AreNotEqual(list, listToTest);
```
breaman: I don't believe this test is doing what you think it is. This is going to be doing a reference equal, which will always be false. You needed to use Enumerable.SequenceEqual.

PR #133, opened 2018-12-04
Assignment9

valdynhunt:


Assignment9/PatentDataAnalyzer/test/PatentDataAnalyzer.Tests/PatentDataAnalyzerTests.cs
```diff
@@ -0,0 +1,86 @@
...
+            IEnumerable<string> lastNames = PatentDataAnalyzer.InventorLastNames();
+            Assert.IsTrue(lastNames.Contains(name));
+        }
+
+
+        [TestMethod]
+        public void InventorLastNames_Returns_List_Reverse_Order_By_Inventor_Id()
+        {
+            IEnumerable<string> lastNames = PatentDataAnalyzer.InventorLastNames();
+            int id = PatentData.Inventors.Length - 1;
```
breaman: This works with this dataset, but only because things are sequential. A more robust way would have been to use PatentData.Inventors.OrderByDesc(i => i.Id).Select(i => i.Name.Split().List()) and make sure those results matched the results from .InventorLastNames().

PR #138, opened 2018-12-04
Bryan Caesar - Assignment 9 Submission

bcaesar42:


Assignment9/src/PatentDataAnalyzer/PatentDataAnalyzer.cs
```diff
@@ -0,0 +1,67 @@
...
+                                        where Inventor.Country.Equals(country)
+                                        select Inventor.Name;
+
+            return query.ToList<string>();
+        }
+
+        public static List<string> InventorLastNames()
+        {
+            IEnumerable<string> query = from Inventor in PatentData.Inventors
+                                        let lastName = Inventor.Name.Split().Last<string>()
```
breaman: Not an issue, but string would be implied here, so you could have just done .Last() instead of .Last<string>()

Assignment9/src/PatentDataAnalyzer/PatentDataAnalyzer.cs
```diff
@@ -0,0 +1,67 @@
...
+                                        orderby Inventor.Id descending
+                                        select lastName;
+
+            return query.ToList<string>();
+        }
+
+        public static string LocationsWithInventors()
```
breaman: This method is not implemented the way the requirements stated. It should have returned a unique list of state-country, where this associates states with wrong countries.
```
return string.Join(", ",
   (from inventor in PatentData.Inventors
   select $"{inventor.State}-{inventor.Country}").Distinct());
```

Assignment9/test/PatentAnalyzer.Tests/PatentAnalyzerTests.cs
```diff
@@ -0,0 +1,95 @@
...
+        [TestMethod]
+        public void Randomize_ResultsDifferFromOriginalAndEachOther()
+        {
+            IEnumerable<string> original = new List<string>() { "Benjamin Franklin", "Orville Wright",
+                                    "Wilbur Wright", "Samuel Morse", "John Michaelis", "Mary Phelps Jacob"};
+
+            IEnumerable<string> random1 = original.Randomize();
+            IEnumerable<string> random2 = original.Randomize();
+
+            Assert.AreNotEqual<IEnumerable<string>>(original, random1);
```
breaman: This test is not testing what you think it is. This is only doing a reference equality, which will always return false. You should have used something like Assert.IsFalse(original.SequenceEquals(random1));

PR #140, opened 2018-12-04
Assignment9

Cwhite534:


Assignment9/Assignment9/Assignment9/PatentDataAnalyzer.cs
```diff
@@ -0,0 +1,61 @@
...
+        {
+            return (from inventor in PatentData.Inventors
+                where inventor.Country == specifiedCountry
+                select inventor.Name).ToList();
+        }
+
+        public static IEnumerable<string> InventorLastNames()
+        {
+            return from inventor in PatentData.Inventors
+                let lastName = inventor.Name.Split(' ').Last()
```
breaman: Nothing wrong with this code, but you could have just called .Split() without the argument and the behavior would have been the same. It splits on whitespace by default.

